PR #7, opened 2019-05-22
systemUI: add adaptive brightness tile

rascarlo:


thestinger:
@rascarlo Maybe you can try submitting this to AOSP? It's unfortunately not the kind of the thing that I'm able to dedicate time to maintaining / porting.

rascarlo:
@thestinger maybe i will. Thanks again for your work

PR #15, opened 2019-09-23
set gestural navigation as default

ghost:


thestinger:
I don't want to make this the default yet for compatibility / consistency reasons. It will probably become the default eventually, but it's a disruptive change that I don't want right now.

PR #18, opened 2019-11-21
require unlocking to use work mode tile

ypid:
Straight forward patch as this is already done for other tiles by @thestinger and for LocationTile.java even from Google itself. Requiring unlocking to use the work mode tile is useful because it would be annoying when someone could just disable your work profile (useful together with Shelter to run apps more isolated than by default). Also, this tile becomes automatically added to the lock screen when you setup a work profile.

Tested on AOSP 10 on sargo. Ref: https://github.com/hashbang/os


thestinger:
Thanks.

ypid:
You are welcome, thanks for merging :)

PR #19, opened 2019-11-24
Don’t require strong auth on regular time interval

ypid:
This disables the need to provide strong authentication except when booting the device. The idea behind this is that you are no longer forced to enter your strong authentication credentials in random locations where it might be easy to snoop your strong authentication credentials allowing an adversary to boot and decrypt your device against your will.

Changing DEFAULT_STRONG_AUTH_TIMEOUT_MS is not enough. Rather, it has the opposite effect. In my tests, it caused the strong auth to be required every hour rather than a 42 d interval.

Ref: frameworks/base/services/devicepolicy/java/com/android/server/devicepolicy/DevicePolicyManagerService.java

In my tests, dpm.getRequiredStrongAuthTimeout(null, userId)) returned 3600000.

Note that this patch circumvents the values that DeviceAdmin may provide. In other words, it ignores whatever a MDM would request for the strong auth timeout!

To be clear, this patch is a practical approach, in an ideal world, we would properly use the DeviceAdmin feature.

Confirmed working on:

* Android 9
* Android 10

Submitted to:

* https://github.com/hashbang/os/pull/32
* https://github.com/RattlesnakeOS/community_patches/pull/9
* https://github.com/ypid/ypid-android-patches

Related to:

* https://github.com/GrapheneOS/os_issue_tracker/issues/160


renlord:
What if the user never reboots their phone and ultimately forgets their strong auth credentials? The regular interval is quite useful at reminding users of their strong auth credentials.

IMO, as a deterrence against snooping, the randomzied keypad/keyboard approach is more promising.

ypid:
I probably should mentioned the context out of which I opened this PR: https://www.reddit.com/r/GrapheneOS/comments/dyjgry/couple_questions_regarding_usage/

As pointed out in https://github.com/hashbang/os/pull/32#issuecomment-523164932, this is of course a downside of this patch. Users must be made aware of this risk (in the case that this PR gets merged, which I am not sure it should). But, even though the 42 d timeout in this patch is not effective, every serious GrapheneOS user will at least reboot every month ;-)

> IMO, as a deterrence against snooping, the randomzied keypad/keyboard approach is more promising.

Not sure on that. In my daily usage, I figured I would like my phone to not ask me for strong auth in any such situation at all. The adversary I imaging could just watch the screen and see exactly what you enter.

thestinger:
I don't want this as an unconditional change since it substantially decreases security compared to the weak authentication mechanism timing out. You need to rethink this.

ypid:
> You need to rethink this.

I agree with you. For reference, I opened https://github.com/linux-colonel/AdminControl/issues/11 where I think such feature could find a home.

PR #27, opened 2020-04-22
Globally (not just in US) opt out of Safe Volume misfeature

rustybird:
Safe Volume has many false alarms. And if overridden by the user, at every 20 hours of accumulated(!) audio playback it will intentionally reset - even while the audio is playing. This can ruin the listening experience, and ironically pose a safety hazard when e.g. the user is operating a vehicle...

AOSP already opted out of Safe Volume in the overlays for Mobile Country Code 310-316 (= US). Make this universal.

### DIFF
--- a/core/res/res/values/config.xml
+++ b/core/res/res/values/config.xml
@@ -2609,7 +2609,7 @@
     <bool name="config_useDevInputEventForAudioJack">false</bool>
 
     <!-- Whether safe headphone volume is enabled or not (country specific). -->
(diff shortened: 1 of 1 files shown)


ypid:
I support this change for exactly the same reasons you mentioned. In fact, I am using it for a good year (ref: https://github.com/ypid/ypid-android-patches/blob/master/patches/disable_safe_media_volume_config.patch). Had no issues with it.

PR #28, opened 2020-04-28
add option to disable fingerprint unlock

### LINKED ISSUE #202: Expose existing support for using fingerprint scanner for apps only in Settings
Fingerprints aren't secure and can easily be bypassed.  Would it be possible like in iOS to disable the fingerprint scanner for unlock but still allow it to be used for apps?

inthewaves:
There is already an implementation for disabling face unlock for the
keyguard, so this commit follows that implementation. (I couldn't find
where the existing code [mentioned here](https://github.com/GrapheneOS/os_issue_tracker/issues/202#issuecomment-612141647) to disable it was, so it's possible that this code
is unnecessary...)

Closes https://github.com/GrapheneOS/os_issue_tracker/issues/202
Review this PR with and find more details at: https://github.com/GrapheneOS/platform_packages_apps_Settings/pull/15

### DIFF
--- a/core/java/android/provider/Settings.java
+++ b/core/java/android/provider/Settings.java
@@ -8325,6 +8325,12 @@ public final class Settings {
         public static final String FLASHLIGHT_ENABLED = "flashlight_enabled";
 
         /**
+         * Whether or not fingerprint unlock is allowed on Keyguard.
+         * @hide
+         */
+        public static final String FINGERPRINT_UNLOCK_KEYGUARD_ENABLED = "fingerprint_unlock_keyguard_enabled";
+
+        /**
          * Whether or not face unlock is allowed on Keyguard.
          * @hide
          */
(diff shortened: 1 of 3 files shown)


core/java/android/provider/Settings.java
```diff
@@ -8323,6 +8323,12 @@ public boolean validate(@Nullable String value) {
          */
         public static final String FLASHLIGHT_ENABLED = "flashlight_enabled";
 
+        /**
+         * Whether or not fingerprint unlock is allowed on Keyguard.
+         * @hide
+         */
+        public static final String FINGERPRINT_UNLOCK_KEYGUARD_ENABLED = "fingerprint_unlock_keyguard_enabled";
```
inthewaves, line 8331: I realized this file contains a `SETTINGS_TO_BACKUP` array (and there's also `VALIDATORS`); should I put this new setting into that array?

inthewaves:
rebased onto QQ3A.200805.001.2020.08.03.22 and retested

inthewaves:
rebased onto QQ3A.200605.002.2020.06.02.02 and retested

core/java/android/provider/Settings.java
thestinger: I'm not sure.

PR #31, opened 2020-08-31
fix VPN lockdown setting getting overridden on user stop

### LINKED ISSUE #213: "Block connections without VPN" option gets unset when multiple users use VPNs
I have five users on a GrapheneOS installation. Each user has a different VPN profile. All of these users will initially have both the options "Always-on VPN" and "Block connections without VPN" enabled.

I've noticed that throughout normal use, right after switching between users, I will briefly see the following notification:
```
Android System
Disconnected from always-on VPN
Change network or VPN settings
```
This notification will disappear after a few seconds. However, once I check the settings for the same VPN profile, the option "Block connections without VPN" will now be disabled, whilst the "Always-on VPN" remains enabled.

This issue does not affect connectivity or even the VPN connection itself. All the various VPN profiles work as expected. The only problem is that the option always gets unset. I can consistently reproduce this issue simply by alternating through the various users. If I switch through each user, one after another, by the time I get back to the start of the cycle, the first user will have had it's "Block connections without VPN" option reset.

I originally believed this to be an issue with my code when working on an Orbot alternative as previously discussed in #210. However, I have reproduced this problem after removing all my VPN profiles and replacing them with new unrelated VPN profiles for the vanilla WireGuard client for Android.

I am using a Pixel 3a on the latest stable build (QQ2A.200501.001.B2.2020.05.05.02).

inthewaves:
Fixes https://github.com/GrapheneOS/os_issue_tracker/issues/213

When a user is stopped/logged out, [this method (Vpn.java)](https://github.com/GrapheneOS/platform_frameworks_base/blob/10/services/core/java/com/android/server/connectivity/Vpn.java#L1328-L1338) is called (doesn't seem to be called on device shutdown/restart, or maybe it is but runs into some exception down below that prevents it from saving the always-on state):
```java
/**
 * Called when the user associated with this VPN has just been stopped.
 */
public synchronized void onUserStopped() {
    // Switch off networking lockdown (if it was enabled)
    setLockdown(false);
    mAlwaysOn = false;

    // Quit any active connections
    agentDisconnect();
}
```
the [`setLockdown`](https://github.com/GrapheneOS/platform_frameworks_base/blob/9300e141fe843876876401fda6beab13d40c78d5/services/core/java/com/android/server/connectivity/Vpn.java#L355-L377) method:
```java
public synchronized void setLockdown(boolean lockdown) {
    enforceControlPermissionOrInternalCaller();

    setVpnForcedLocked(lockdown);
    mLockdown = lockdown;

    // Update app lockdown setting if it changed. Legacy VPN lockdown status is controlled by
    // LockdownVpnTracker.isEnabled() which keeps track of its own state.
    if (mAlwaysOn) {
        saveAlwaysOnPackage();
    }
}
```
the [`saveAlwaysOnPackage`](https://github.com/GrapheneOS/platform_frameworks_base/blob/9300e141fe843876876401fda6beab13d40c78d5/services/core/java/com/android/server/connectivity/Vpn.java#L550-L567) method
```java
/**
 * Save the always-on package and lockdown config into Settings.Secure
 */
@GuardedBy("this")
private void saveAlwaysOnPackage() {
    final long token = Binder.clearCallingIdentity();
    try {
        mSystemServices.settingsSecurePutStringForUser(Settings.Secure.ALWAYS_ON_VPN_APP,
                getAlwaysOnPackage(), mUserHandle);
        mSystemServices.settingsSecurePutIntForUser(Settings.Secure.ALWAYS_ON_VPN_LOCKDOWN,
                (mAlwaysOn && mLockdown ? 1 : 0), mUserHandle);
        mSystemServices.settingsSecurePutStringForUser(
                Settings.Secure.ALWAYS_ON_VPN_LOCKDOWN_WHITELIST,
                String.join(",", mLockdownWhitelist), mUserHandle);
    } finally {
        Binder.restoreCallingIdentity(token);
    }
}
```

So, when a user is stopped, it seems to always put in `0` for the user's `Settings.Secure.ALWAYS_ON_VPN_LOCKDOWN`, because `mLockdown` is false when the always-on package is saved. And the always on package is saved because `setLockdown` is called when mAlwaysOn is true.

So we can just overload the method for compatibility, moving the original in a private method to handle the save state boolean parameter.

Tested on Pixel 3a with WireGuard and OpenVPN, including testing that the lockdown setting is saved on logout (by switching 3 profiles to get least recently used user logged out), internet not affected after forgetting VPN, lockdown actually works, the right VPN is remembered as the always-on VPN, other profiles are unaffected, VPN settings remembered after restarting device, forgetting VPN still works

### DIFF
--- a/services/core/java/com/android/server/connectivity/Vpn.java
+++ b/services/core/java/com/android/server/connectivity/Vpn.java
@@ -416,6 +416,22 @@ public class Vpn {
      * @param lockdown whether to prevent all traffic outside of a VPN.
      */
     public synchronized void setLockdown(boolean lockdown) {
+        // Save the always-on state, since this is expected to be called
+        // for changing the lockdown/always-on setting.
+        setLockdown(lockdown, true);
+    }
+
+    /**
+     * Does the same as {@link #setLockdown(boolean)}, but allow for prevention of saving the
+     * always-on state. When users are stopped, lockdown is turned off; this originally
+     * called saveAlwaysOnPackage with mLockdown == false into Settings.Secure, overriding
+     * what was saved before. Therefore, we need to provide a way to not call setAlwaysOnPackage
+     * for that case (GrapheneOS/os_issue_tracker#213).
+     *
+     * @param lockdown whether to prevent all traffic outside of a VPN.
+     * @param saveAlwaysOnState whether to save the always-on state.
+     */
+    private synchronized void setLockdown(boolean lockdown, boolean saveAlwaysOnState) {
         enforceControlPermissionOrInternalCaller();
 
         setVpnForcedLocked(lockdown);
@@ -423,7 +439,7 @@ public class Vpn {
 
         // Update app lockdown setting if it changed. Legacy VPN lockdown status is controlled by
         // LockdownVpnTracker.isEnabled() which keeps track of its own state.
-        if (mAlwaysOn) {
+        if (mAlwaysOn && saveAlwaysOnState) {
             saveAlwaysOnPackage();
         }
     }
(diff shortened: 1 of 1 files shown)


inthewaves:
Submitted a similar fix upstream; might just cherry pick that if it goes through

thestinger:
@inthewaves Can you move this over to the Android 11 branch?

inthewaves:
@thestinger I cherry-picked it, then built with and without this commit to test in another user for Android 11. Seems to fix it

PR #33, opened 2020-09-02
show hidden logout button and allow system user logout

### LINKED ISSUE #258: Add logout button to logout of current user profile
https://freenode.logbot.info/grapheneos/20200709#c4343317

> you can also take advantage of profiles already
>
> put sensitive stuff in a profile with a strong passphrase
>
> use a profile with a short PIN for quick use
>
> keep the sensitive profile inactive (logged out)
>
> atm there isn't an explicit logout option in the GUI but you can reboot for now, 
> we could easily add a logout to the power menu
>
> just another thing people have to step up to help implement
>
> logout / purging the keys is already supported
>
> just isn't a button for it in the GUI

inthewaves:
Fixes https://github.com/GrapheneOS/os_issue_tracker/issues/258

Exposes the logout button in the power menu already built into the operating system and allows for logging out of the system/admin user. This logout button should evict the credential keys for the user. If the user being logged out is not the system/admin user, it will switch to the system user and then stop the user.

Logging out of the system user is not originally supported by the operating system as seen from all these checks in UserController. 
- Instead of modifying the original logic (there are a lot of checks to make sure the system user isn't being logged out), new methods that follow the original methods are used to logout of the system user. This is done to keep things compatible with previous code, and to only allow stopping the system user when the user explicitly presses the logout button.
- A KeyEvictedCallback is used to restart the system user after it has been stopped and cleaned up. If we don't do it that way, then the screen will be blank after the system user stops.

### DIFF
--- a/core/java/android/app/IActivityManager.aidl
+++ b/core/java/android/app/IActivityManager.aidl
@@ -344,6 +344,8 @@ interface IActivityManager {
     @UnsupportedAppUsage
     int stopUser(int userid, boolean force, in IStopUserCallback callback);
     @UnsupportedAppUsage
+    int restartSystemUserInForegroundIfCurrent();
+    @UnsupportedAppUsage
     void registerUserSwitchObserver(in IUserSwitchObserver observer, in String name);
     void unregisterUserSwitchObserver(in IUserSwitchObserver observer);
     int[] getRunningUserIds();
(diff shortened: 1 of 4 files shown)


inthewaves:
Left as a draft for now for anyone else to look at due to the possible concerns over data still visible in plaintext when viewed in adb shell with root (https://github.com/GrapheneOS/os_issue_tracker/issues/258#issuecomment-684677223). Might be something with how adb shell works, don't know. The data is not in plaintext when the phone is restarted though, and some logouts don't show any plaintext files in `/data/user/<some userId>`

PR #34, opened 2020-09-10
require unlocking to use sensitive quick tiles

inthewaves:
**Edit**: This now just ports all the sensitive tiles. No longer using KeyguardDismissUtil. See https://github.com/GrapheneOS/platform_frameworks_base/pull/34#issuecomment-690808872

Closes https://github.com/GrapheneOS/os_issue_tracker/issues/277

---

Old version from 10:
https://github.com/GrapheneOS/platform_frameworks_base/commit/73275059b13c84ca0844c1dcc0bcc46db8f005f2

This version follows the KeyguardDismissUtil approach used in the new screen record tile (`showPrompt` is posted to the UI handler): https://github.com/GrapheneOS/platform_frameworks_base/blob/231ac9bcedb0133720729f9d73fe9667a5580264/packages/SystemUI/src/com/android/systemui/qs/tiles/ScreenRecordTile.java#L115-L124

The `requiresShadeOpen` parameter should open the panels like the version for Android 10

Wanted to run this for a review before doing this for all the other sensitive tiles

### DIFF
--- a/packages/SystemUI/src/com/android/systemui/qs/tiles/AirplaneModeTile.java
+++ b/packages/SystemUI/src/com/android/systemui/qs/tiles/AirplaneModeTile.java
@@ -38,6 +38,7 @@ import com.android.systemui.plugins.qs.QSTile.BooleanState;
 import com.android.systemui.qs.GlobalSetting;
 import com.android.systemui.qs.QSHost;
 import com.android.systemui.qs.tileimpl.QSTileImpl;
+import com.android.systemui.statusbar.policy.KeyguardStateController;
 
 import javax.inject.Inject;
 
@@ -48,11 +49,14 @@ public class AirplaneModeTile extends QSTileImpl<BooleanState> {
     private final ActivityStarter mActivityStarter;
     private final BroadcastDispatcher mBroadcastDispatcher;
 
+    private final KeyguardStateController mKeyguard;
+
     private boolean mListening;
 
     @Inject
     public AirplaneModeTile(QSHost host, ActivityStarter activityStarter,
-            BroadcastDispatcher broadcastDispatcher) {
+            BroadcastDispatcher broadcastDispatcher,
+            KeyguardStateController keyguardStateController) {
         super(host);
         mActivityStarter = activityStarter;
         mBroadcastDispatcher = broadcastDispatcher;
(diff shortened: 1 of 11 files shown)


inthewaves:
Actually, I'm going to do all of them at once

inthewaves:
There's an upstream bug with this implementation when using `true` for the `requiresShadeOpen` parameter in the `KeyguardDismissUtil.executeWhenUnlocked` method.

It seems the bug comes from `mLeaveOpenOnKeyguardHide` inside of StatusBarStateControllerImpl.java, The bug seems to be that `mLeaveOpenOnKeyguardHide` is left at `true` when the status bar is closed, resulting in some weird behavior after the keyguard is hidden (probably any code that reads that boolean field expects the status bar to still be open: "leaveOpen").

`KeyguardDismissUtil.executeWhenUnlocked`, `StatusBar.postQSRunnableDismissingKeyguard` and the lockscreen notification expansion gesture code can set `mLeaveOpenOnKeyguardHide` to true, leading to this bug. Able to reproduce the bug via the
notification gesture on the stock OS (sargo-rp1a.200720.009). See the GIFs below for more details. Was not able to reproduce this via the notification gesture on a Android 10 (GrapheneOS QQ3A.200805.001.2020.08.07.01) release.

How to reproduce for the tiles:
1. tapping any sensitive tile on the lockscreen,
2. pressing the back button to close the keyguard input,
3. collapsing the notification panel completely, and waiting for a few
   seconds,
4. unlocking the lock screen to hide keyguard.

Then, the panel bar no longer becomes expandable until you keep trying.

Will attempt a fix/workaround the bug by setting mLeaveOpenOnKeyguardHide to be false when the status bar panel is closed.

inthewaves:
Left: Current GrapheneOS 11 branch
Right: Stock OS, sargo-rp1a.200720.009

|pull-down-bug-graphene20 | pull-down-bug-stock20 |
|---|---|

thestinger:
I'll wait until all the commits are ported before merging this. Not sure what to do about that bug.

inthewaves:
@thestinger Should be ready now. I can't change the originating branch on GitHub, so I just overwrote the port-wifi-tile branch

Implementation-wise, I moved away from KeyguardDismissUtil and went back over to what the Location tile is using to enforce keyguard. Using that made it inconsistent with how it acted with the Location tile

Bug-wise, I tried to implement a fix of it in https://github.com/GrapheneOS/platform_frameworks_base/pull/34/commits/bb9c137b2a69a9fb000b774ac0eabda7b43847c9. It seems to be some Android 11 specific bug, as I'm not able to reproduce it on an Android 10 release.

thestinger:
Should report the bug upstream via the bug tracker:

https://source.android.com/setup/contribute/report-bugs#bug-queues

I guess it falls under framework? Their categories are... odd.

inthewaves:
Tracking at https://issuetracker.google.com/issues/168300271

PR #42, opened 2020-10-04
support upgrading GrapheneOS-specific settings, and fix Seedvault SettingsProvider code not running

inthewaves:
Related to https://github.com/GrapheneOS/os_issue_tracker/issues/261, though no defaults were added yet.

This is done to reduce the maintenance burden needed to deal with upstream settings migrations and to also allow us to do any upgrades / migrations with GrapheneOS-specific settings. (e.g., See https://github.com/GrapheneOS/platform_frameworks_base/commit/10282ec0dead9a83abc9623a139fa45ed67894a7, where the added Seedvault-related code in SettingsProvider right now never runs, because currentVersion != 183 at that point of the code.) It's only run if the version changes, when new users are added---basically any other time a normal setting upgrade would occur.

When some upgrade is needed (e.g., setting defaults reliably, or moving from Secure to Global or vice versa), add the upgrade code in onUpgradeGrapheneOsSettingsLocked, and update the GRAPHENEOS_SETTINGS_VERSION field in SettingsProvider#UpdateController.

Upgrading of GrapheneOS-settings uses an independent version from the normal settings version, thus upgrades are done independently (but done after the normal settings upgrades). This means that we don't have to worry as much about managing conflicts between GrapheneOS settings and settings version upgrades from upstream.

For now, the GrapheneOS settings version is stored in Settings.Secure, even though the normal settings version is stored as an XML attribute.

The SettingsProvider changes from https://github.com/GrapheneOS/platform_frameworks_base/commit/10282ec0dead9a83abc9623a139fa45ed67894a7 were moved into this new upgrade system.

---

If this change makes it into a release, a GrapheneOS setting upgrade will occur for all users, since they wouldn't have a version for it yet. This means the SettingsProvider code from https://github.com/GrapheneOS/platform_frameworks_base/commit/10282ec0dead9a83abc9623a139fa45ed67894a7 is rerun again. However, rerunning the Seedvault code in SettingsProvider shouldn't do anything, since backups are expected to always be with Seedvault.


inthewaves:
Could be used in the future to set all current users to be disallowed from running in the background via setting something like UserManager.DISALLOW_RUN_IN_BACKGROUND

thestinger:
This makes sense but I've just dropped the Seedvault migration code since ample time has been provided for users to upgrade already.

thestinger:
This will be considered again later.

PR #45, opened 2020-10-05
prevent more unchecked null pointers stored in a non-null Drawable type

inthewaves:


inthewaves:
the same file as https://github.com/GrapheneOS/platform_frameworks_base/pull/43/commits/9b3150a2c1c6d2e6a8c0378d45fae75a3c31147b

PR #50, opened 2020-11-27
microG Support

dylangerdaly:
Allows microG Spoofing, matching on the sha256 hash of the signing certificate and package ID.


services/core/java/com/android/server/pm/PackageManagerService.java
```diff
@@ -703,6 +703,12 @@
 
     private static final String RANDOM_DIR_PREFIX = "~~";
 
+    /**
+     * The Google signature faked by microG.
```
thestinger, line 707: Instead of faked "by microG", can you say "for microG" to make it clear that it's not microG doing it?

core/res/res/values/config.xml
```diff
@@ -1646,6 +1646,8 @@
     <string-array name="config_locationProviderPackageNames" translatable="false">
         <!-- The standard AOSP fused location provider -->
         <item>com.android.location.fused</item>
+        <!-- The (faked) microg fused location provider (a free reimplementation) -->
```
thestinger, line 1649: We should start without this and consider it later on. I want to start without granting it any special privileges, and we want to provide supplementary location services ourselves within GrapheneOS.

services/core/java/com/android/server/pm/PackageManagerService.java
```diff
@@ -4432,6 +4439,24 @@ private PackageInfo generatePackageInfo(PackageSetting ps, int flags, int userId
         }
     }
 
+    private PackageInfo mayFakeSignature(AndroidPackage p, PackageInfo pi,
+            Set<String> permissions) {
+    	String hash = p.getSigningDetails().getSha256Certificate();
+        try {
+            if (hash.equals(microGHash)
+                    && p.getTargetSdkVersion() > Build.VERSION_CODES.LOLLIPOP_MR1) {
```
thestinger, line 4447: You might as well update this to the current target SDK version used by microG. It's hopefully higher than Lollipop.

thestinger: Or, as @anupritaisno1 suggest, just drop this because we'll be specifically checking for the app signed with our own signing key and we won't ever have one without a modern target version anyway.

renlord:
@thestinger, what else is needed for this PR to be accepted?

SamirHafez:
@dylangerdaly I thought the plan for microG support was to provide a base that would allow Google Play implementations to be plugged in without having to give it spoofing permissions (or root like powers). Then users could choose and install an implementation (microG, stub, etc) and have it handle Google Play.

Is this effort going towards that or was my assumption incorrect about the direction of the proposed solution?

I ask because I'd like to understand what the agreed way forward is here so I can help where possible.

dylangerdaly:
Hey @SamirHafez, indeed this patch removes the permissions and only allows based on the hash of the certificate used to sign microG, or a stub etc

This hash is baked into the OS and will be controlled by GrapheneOS (Graphene will compile and sign it's own version of microG)

I believe Graphene is working out how to distribute the apk and that's why this has been on pause

thestinger:
I'm going to close this until a full implementation of what's needed is available. I don't think we're going to be able to fork and maintain microG so the approach needs to be implementing a stub implementation of Play services pretending the services are unavailable.

kalbasit:
> I'm going to close this until a full implementation of what's needed is available. I don't think we're going to be able to fork and maintain microG so the approach needs to be implementing a stub implementation of Play services pretending the services are unavailable.

How long do you estimate that's going to take? I'm currently blocked on the full migration to Graphene.

thestinger:
https://grapheneos.org/#never-google-services has our current plans. We're only going to be taking the approach of making a stubbed out implementation of Play services after realizing the scale of what it would take to turn microG into something acceptable.

dm17:
> https://grapheneos.org/#never-google-services has our current plans. We're only going to be taking the approach of making a stubbed out implementation of Play services after realizing the scale of what it would take to turn microG into something acceptable.

Thanks, is there a thread for the "stubbed out" implementation yet? Hoping this will allow me to launch the Uber app eventually (I also found various other ride sharing apps depended on Google Services when I tried them).

thestinger:
You can use https://m.uber.com/.

dm17:
> You can use https://m.uber.com/.

You can't do change destination or do multiple stops. It has other pain points too.

thestinger:
Their app heavily uses Play services and it's not going to work without it. They use Google services for payments, etc. via Play services. It's not going to work with a stubbed out implementation. It doesn't work with microG.

If you want it to work, then become part of a movement to get app developers to make their apps work without Play services.

dm17:
> Their app heavily uses Play services and it's not going to work without it. They use Google services for payments, etc. via Play services. It's not going to work with a stubbed out implementation. It doesn't work with microG.
> 
> If you want it to work, then become part of a movement to get app developers to make their apps work without Play services.

Yes, I'm pushing from that direction as well :)
Thanks, good to know.

PR #52, opened 2020-12-29
Changed Captive Portal URL From Google To GrapheneOS

ghost:


thestinger:
What I meant is that this is only updating a test case, and the commit message should reflect that only a test case is being changed.

thestinger:
The captive portal URL is set in `services/core/java/com/android/server/ConnectivityService.java`. This isn't changing the URL. It's changing a test case run as part of the test suite.

ghost:
You can close this pull & free to commit as your wish :smile:

thestinger:
This is what I meant:

https://github.com/GrapheneOS/platform_frameworks_base/commit/34391cc0c2906031e777bbc0e6175062d7a7ea8c

ghost:
BTW thanks for the commit :smile:

PR #54, opened 2021-01-08
Use true black for Settings background

ghost:


Exilion666:
> Use true black for Settings background when using the Dark theme. This
> has no effect on the Light theme.
> 
> Fixes [GrapheneOS/os_issue_tracker#355](https://github.com/GrapheneOS/os_issue_tracker/issues/355)

Could you please take a look if it possible to change colour scheme of the dialler app / enable to follow dark mode? In lineage Os and CalyxOS it works. Thanks.

thestinger:
It's not relevant to that.

PR #59, opened 2021-03-16
Remove DUN requirement for tethering

flawedworld:
DUN is what carriers use to identify if traffic is being sent via tethering. From my understanding this commit will prevent carriers from identifying tethering traffic.


flawedworld:
See https://github.com/GrapheneOS/os_issue_tracker/issues/70

PR #61, opened 2021-04-25
Basic Fix for INERNET Runtime permission working owner profile permission instead of users profile

empratyush:


empratyush:
getPermissionForUid(uid) was returning main/owner profile permission status instead of user specific. Fixed By Extracting user id from UID and querying package info as user.

In enforceINTERNETAsRuntimePermission method "getPackageUid" is marked as deprecated and suggested to use getPackageUidInternal, so changes to that.

 getPackageInfo was also returning main/owner profile permission status instead of user specific. Fixed By passing user id from calling method and querying package info as user.

PR #68, opened 2021-06-04
automatically reboot device after timeout if set

ghost:
Signed-off-by: anupritaisno1 <www.anuprita804@gmail.com>
Change-Id: I4a8223a34e332d28b61f5e59440a78f958de8304


ghost:
Why reboot vs halt?

thestinger:
Reboot is friendlier and has no disadvantages for the use case.

ghost:
What is the use case, curiously? Evacuating keys out of memory? Lockdown should be able to do that.

If device is un-handled after X hours, some would prefer to save battery by keeping the device off. I think you can only have a scheduled alarm function until the device is unlocked?

ghost:
lockdown does NOT "evacuate" keys from memory

ghost:
this needs further testing. it seems it is not as strict as the requirements for the feature intended it to be. pushed a small patch that maybe fixes it

thestinger:
@msoose Please read https://grapheneos.org/faq#encryption. You're confusing lockdown (disable secondary unlock) with end session (which is offered by GrapheneOS but not the stock OS / AOSP). End session is not a substitute for this feature. End session is manual, not automatic, and only applies to 1 session. The whole point of this feature is that it's automatic, not manual.

thestinger:
This is a security feature for putting data at rest, not a way to preserve the battery of an idle device that's not charging. I don't know why you're bringing up the manual lockdown action, even if it was a way to clear keys, which it isn't. That's end session (logout).

ghost:
above commit should fix the initial bugs discovered by testers. still needs more testing

ghost:
Thank you for posting the link which precisely explains the differences between the 2. Makes complete sense.

I understand it's a security feature, and the specific feature it provides, but is rebooting 'friendlier' vs simply remaining in an off state, where all radios are supposedly also turned off?

thestinger:
I don't think you're understanding the point of the feature and your comments are starting to come across as concern trolling.

thestinger:
I don't think you're understanding the point of the feature and your comments are starting to come across as concern trolling.

ghost:
fixed an issue i discovered myself. it seems while reimplementing the cancellation code i put the cancel() call in a function that wasn't the right one

this force push moves the cancellation function call to the correct place

ghost:
fixed formatting

packages/SystemUI/src/com/android/systemui/keyguard/KeyguardViewMediator.java
```diff
@@ -1409,6 +1433,10 @@ private void doKeyguardLocked(Bundle options) {
 
         if (DEBUG) Log.d(TAG, "doKeyguard: showing the lock screen");
         showLocked(options);
+        final int rebootAfterTimeout = Settings.Global.getInt(mContext.getContentResolver(), "settings_reboot_after_timeout", 0);
```
inthewaves: 1. It would be better if the key `"settings_reboot_after_timeout"` was stored in [Settings.Global](https://github.com/GrapheneOS/platform_frameworks_base/blob/11/core/java/android/provider/Settings.java#L9109), like what we've done for other settings
2. Using `Settings.Global.getLong` can make this more futureproof for longer intervals

packages/SystemUI/src/com/android/systemui/keyguard/KeyguardViewMediator.java
```diff
@@ -1058,6 +1078,10 @@ private void cancelDoKeyguardForChildProfilesLocked() {
         mDelayedProfileShowingSequence++;
     }
 
+    private void cancelDoRebootForOwnerAfterTimeoutIfEnabled() {
+        mDelayedRebootSequence++;
```
inthewaves, line 1082: can we also call `mAlarmManager.cancel` on the PendingIntent here? might need to refactor the PendingIntent creation from `doRebootForOwnerAfterTimeoutIfEnabled` so that a PendingIntent can be created for here too

packages/SystemUI/src/com/android/systemui/keyguard/KeyguardViewMediator.java
```diff
@@ -1409,6 +1433,10 @@ private void doKeyguardLocked(Bundle options) {
 
         if (DEBUG) Log.d(TAG, "doKeyguard: showing the lock screen");
         showLocked(options);
+        final int rebootAfterTimeout = Settings.Global.getInt(mContext.getContentResolver(), "settings_reboot_after_timeout", 0);
+        if (rebootAfterTimeout != 0) {
+            doRebootForOwnerAfterTimeoutIfEnabled(rebootAfterTimeout);
+        }
```
inthewaves, line 1439: Because of the nature of this feature, it feels like it would be good if we hardcoded a lower bound like 1 hour  (which is the lowest value of the UI implementation)

inthewaves (changes requested):

test-acc-vaccym:
It should not restart before first unlock

packages/SystemUI/src/com/android/systemui/keyguard/KeyguardViewMediator.java
ghost: 2. i do not see the point of a reboot timeout longer than a month

packages/SystemUI/src/com/android/systemui/keyguard/KeyguardViewMediator.java
ghost: i do not want to hardcode a default value. this feature is disabled by default

packages/SystemUI/src/com/android/systemui/keyguard/KeyguardViewMediator.java
ghost: ok i see now, you want the key to be stored in that file. i will move it there

packages/SystemUI/src/com/android/systemui/keyguard/KeyguardViewMediator.java
ghost: unsure how to do that. can you suggest something?

packages/SystemUI/src/com/android/systemui/keyguard/KeyguardViewMediator.java
inthewaves: I'm not talking about a default value; I'm talking about a hardcoded guard, like `if (rebootAfterTimeout >= 1 hour)`. This will maintain that the feature is disabled by default, but just ensures that it's not set to some very low time somehow

packages/SystemUI/src/com/android/systemui/keyguard/KeyguardViewMediator.java
inthewaves: Extract this part into a method that returns a PendingIntent:

```
private PendingIntent getRebootPendingIntent() {
    Intent rebootIntent = new Intent(DELAYED_REBOOT_ACTION);
    rebootIntent.putExtra("seq", mDelayedRebootSequence);
    rebootIntent.addFlags(Intent.FLAG_RECEIVER_FOREGROUND);
    return PendingIntent.getBroadcast(mContext, 0, rebootIntent, PendingIntent.FLAG_CANCEL_CURRENT);
}
```

Then, the `getRebootPendingIntent` method can be used in two places: to set the alarm and to cancel the alarm (https://developer.android.com/reference/android/app/AlarmManager#cancel(android.app.PendingIntent))

https://github.com/GrapheneOS/platform_frameworks_base/blob/bad07fdd0b4528ef166f0e9e692be641ac1ae99e/core/java/android/app/AlarmManager.java#L905-L915

[`Intent#filterEquals` doesn't compare the extra data](https://github.com/GrapheneOS/platform_frameworks_base/blob/bad07fdd0b4528ef166f0e9e692be641ac1ae99e/core/java/android/content/Intent.java#L10331-L10360), so in `cancelDoRebootForOwnerAfterTimeoutIfEnabled()`, we can just get the PendingIntent and then call `mAlarmManager.cancel` on it

packages/SystemUI/src/com/android/systemui/keyguard/KeyguardViewMediator.java
ghost: we don't really support users changing these values through adb or other unsupported methods. that would be completely on them if they messed up

ghost: anyway it seems 2 minutes is good enough for the guard value

packages/SystemUI/src/com/android/systemui/keyguard/KeyguardViewMediator.java
ghost: this would be something to consider for later

thestinger:
This is landed.

PR #71, opened 2021-07-01
Bluetooth auto turn off

empratyush:
#[issue 528](https://github.com/GrapheneOS/os-issue-tracker/issues/528) can be closed.


empratyush:
previous implementation had a lot of flaw. SystemUi is not counted a system App thus it can't send privilege broadcast. 

Settings app is supposed to change system preferences and should NOT be an actual feature implementation.

previous changes need to be dropped.

PR #87, opened 2021-08-08
SUPL: Don't send IMSI / Phone number to SUPL server

SkewedZeppelin:
From https://review.lineageos.org/c/LineageOS/android_frameworks_base/+/249219

Tested working on 7-11 for well over a year now.


thestinger:
This should be clarified since Google's SUPL server is not the one being used by a lot of carriers.

ghost:
The last time I tested this patch, it completely broke network connectivity checks for a few applications

Has that been fixed?

SkewedZeppelin:
> This should be clarified

I've adjusted the wording.

> broke network connectivity checks

Have not encountered this.

thestinger:
So the issue I see with this simply being mandatory is that some carriers may actually require sending this information even though Google's SUPL server doesn't require it. It might need to be a toggle. We could potentially do a Beta release to get people to test it and see if anyone has issues.

thestinger:
I don't see how it could break connectivity checks.

thestinger:
This will need to be ported to 12.

PR #91, opened 2021-08-31
Configurable max biometric attempts

flawedworld:


flawedworld:
This needs the relevant GUI in Settings to be sorted out and face needs to be tested

PR #111, opened 2021-10-30
SUPL: Don't send IMSI / Phone number to SUPL server

SkewedZeppelin:
From https://review.lineageos.org/c/LineageOS/android_frameworks_base/+/249219

Tested working on 7-11 for well over a year now.

Previously:
https://github.com/GrapheneOS/platform_frameworks_base/pull/87


services/core/java/com/android/server/location/gnss/GnssLocationProvider.java
```diff
@@ -1530,6 +1530,11 @@ public void onRequestSetID(@GnssNative.AGpsCallbacks.AgpsSetIdFlags int flags) {
         int type = AGPS_SETID_TYPE_NONE;
         String setId = null;
 
+        /*
```
thestinger, line 1533: SUPL very noticeably improves GPS so this comment is out-of-place. If this breaks SUPL from working properly, it needs to be a configuration option. If it works fine with every major carrier without including these, then it makes sense to unconditionally disable them this way. We do care about A-GPS working particularly since we don't have network location services. It's mandatory to keep that working and we can only merge this if it doesn't break it. Comment implies that this breaks it and downgrades it to not having A-GPS which would be a serious problem.

thestinger: So, please remove the whole comment and make sure that SUPL / A-GPS still works properly with this change. We'll need to properly test it across multiple carriers in order to include it. The comment is currently giving a bad impression of the change and is part of why it has taken so long to consider moving forward with it.

thestinger: Also, just delete the code instead of commenting it out. It should either be deleted or made conditional based on a system property controlled by a setting if it has a real use case. It might be better to make the change elsewhere though.

PR #113, opened 2021-11-02
Change settings' dark theme color to pure black

quh4gko8:
Based on this commit at last tag at R: https://github.com/GrapheneOS/platform_frameworks_base/commit/959b83650ec705b8a2c5ebdb0bba9aa8a2086d91
but using the colors.xml black value at L39
https://github.com/GrapheneOS/platform_frameworks_base/blob/12/core/res/res/values/colors.xml#L39


thestinger:
I think we want to stick with how this already works.

PR #115, opened 2021-11-16
gmscompat: improve Play Store compatibility

muhomorr:
What works:
- package (un)installation confirmation
- unattended app upgrades (for apps that target API 29+)

What doesn't work:
- Play Store self-update. Should work once it starts to use the modern APIs.
- Play Store doesn't expect package uninstallation to fail, 
progress spinner continues indefinitely if confirmation is declined.

`UPDATE_PACKAGES_WITHOUT_USER_ACTION` and `REQUEST_DELETE_PACKAGES` permissions should be granted to Play Store during the OS upgrade. To grant them immediately, reinstall the Play Store app.

### DIFF
--- a/core/java/android/app/ApplicationPackageManager.java
+++ b/core/java/android/app/ApplicationPackageManager.java
@@ -122,6 +122,7 @@ import com.android.internal.annotations.GuardedBy;
 import com.android.internal.annotations.Immutable;
 import com.android.internal.annotations.VisibleForTesting;
 import com.android.internal.gmscompat.GmsHooks;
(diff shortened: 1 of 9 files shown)


thestinger:
> Play Store self-update. Should work once it starts to use the modern APIs.

Is it using some older version of the privileged API for this?

muhomorr:
I have overestimated the amount of effort necessary to make self-update work:
https://github.com/GrapheneOS/platform_frameworks_base/pull/116
https://github.com/GrapheneOS/platform_packages_providers_DownloadProvider/pull/2

PR #118, opened 2021-11-23
gmscompat: fix push notifications in secondary user profiles

### LINKED ISSUE #734: Newer versions (21.33.14 and above) of gms break push notifications in secondary user profile
Device: Pixel 5
OS version: SP1A.210812.015.2021102000

This issue seems unrelated to the OS version, and to me has existed since gms 21.33.14 came out (around 9/23). After updating to this version, Signal and other apps using FCM could not receive push notifications anymore. Completely removing and reinstalling gsf, gms and vending apps did not help, but installing older version of gms (21.30.16) did.  I think google changed something between these two versions of gms. 
In owner profile FCM works fine with all versions of gms. The issue only exists in secondary profile. Versions of gsf and vending apps seem irrelevant. 

Tested not working: 
gms 	21.33.14 (150400-395723304) , 21.39.16 (190400-402663742)

Tested working: 
gms		21.30.16 (150400-391784508) , 21.30.16 (190400-391784508)    (and all previous versions since I hadn't had any problems till 21.33.14)

muhomorr:
`UserManager.isUserUnlocked(userId)` returns whether a user exited the Direct Boot mode.
This method requires a restricted permission if userId parameter differs from the current userId.
GMS called this method with userId 0 in both primary (`userId == 0`) and secondary (`userId != 0`)
profiles. userId was always 0 likely because of the stubbed out
`UserManager.getUserSerialNumber(userId)`,
which always returned 0 as a workaround to a different issue.

Tested with [Slack's "Troubleshoot Notifications" tool](https://slack.com/help/articles/360001559367-Troubleshoot-Slack-notifications#android-2).

Closes https://github.com/GrapheneOS/os-issue-tracker/issues/734

### DIFF
--- a/core/java/android/os/UserManager.java
+++ b/core/java/android/os/UserManager.java
@@ -2527,6 +2527,9 @@ public class UserManager {
     @RequiresPermission(anyOf = {Manifest.permission.MANAGE_USERS,
             Manifest.permission.INTERACT_ACROSS_USERS}, conditional = true)
     public boolean isUserUnlocked(@UserIdInt int userId) {
+        if (GmsCompat.isEnabled()) {
+            return true;
+        }
         return mIsUserUnlockedCache.query(userId);
     }
 


thestinger:
Instead of hard-wiring it to `true`, this should likely replace the userid with the current userid since we want Play services to be able to get the result but we want it to use secondary users as the main user.

PR #120, opened 2021-12-08
gmscompat: set foregroundServiceType of location services

### LINKED ISSUE #635: Location stops working with Sandboxed Play Services installed
**My situation and what I've done:**
- Running the stable build RQ3A.210805.001.A1.2021081822 on my Pixel 4a
- Successfully installed Sandboxed Play Services like mentioned in this guide: https://grapheneos.org/usage#sandboxed-play-services
- Enabled location permission for Google Play Services
- Checked that location is enabled correctly for each app which is requesting the permissions
- Tried multiple apps which need location services to navigate (Magic Earth, Osmand, Google Maps)

**Issue:**
My location only seems to be recognized correctly after a device reboot by several apps.
When the device is up for around 3-4 minutes, the location stops to be working until the next reboot of my device.

**Different behavior by different apps:**
- Magic Earth: After reboot working immediately, after closing the app and waiting 3-4 minutes location is still shown but stuck
- Osmand: After reboot working immediately, after closing the app and waiting 3-4 minutes no location is found until reboot
- Google Maps: After reboot working immediately, after closing the app and waiting 3-4 minutes no location is found until reboot

When location permission for Google Play Services is removed, none of these apps find my location. Not even after a reboot.

This issue seems only to occur when Sandboxed Play Services are installed.

muhomorr:
Without it, location access gets throttled after a few minutes.

Make sure to reinstall GMS so that its AndroidManifest is reparsed (it also gets reparsed during the OS upgrade).

Closes https://github.com/GrapheneOS/os-issue-tracker/issues/635

### DIFF
--- a/core/java/android/content/pm/parsing/component/ParsedServiceUtils.java
+++ b/core/java/android/content/pm/parsing/component/ParsedServiceUtils.java
@@ -32,6 +32,7 @@ import android.content.res.XmlResourceParser;
 import android.os.Build;
 
 import com.android.internal.R;
+import com.android.internal.gmscompat.GmsInfo;
 
 import org.xmlpull.v1.XmlPullParser;
 import org.xmlpull.v1.XmlPullParserException;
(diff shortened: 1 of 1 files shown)


muhomorr:
Needs more testing.

thestinger:
Applied this so we can try it out via the Beta channel and see if it works for people.

PR #126, opened 2021-12-18
Show notification if FIPS is enabled on GS101 devices

flawedworld:
FIPS mode on GS101 devices disables wrapped key encryption, and decreases device security, whilst this itself does not compromise the integrity of the device, GrapheneOS does not support this configuration. Alert the user to reinstall GrapheneOS to avoid potential data loss in future.


thestinger:
We're now ignoring this so it shouldn't be necessary anymore unless this has an impact beyond what we disabled.

PR #130, opened 2022-01-08
Set com.android.imsserviceentitlement/.WfcActivationActivity as defau…

flawedworld:
…lt emergency address app for Wi-Fi calling

This will not work on all carriers but it's the best we can do.


flawedworld:
Debate to be had over if this is the best choice or not.

flawedworld:
ignore, do not merge

PR #141, opened 2022-02-04
Active edge

flawedworld:


thestinger:
Not currently planned unless Pixel 7 adds back active edge.

PR #167, opened 2022-03-23
Spoof build fingerprint for Google Play Services

flawedworld:
SafetyNet's CTS profile attestation checks whether Build.FINGERPRINT
matches that of the device's stock OS, which has passed CTS testing.
Spoof the fingerprint for Google Play Services to help pass SafetyNet.

We used to set the real system build fingerprint to the stock one, but
Android relies on each build having a unique fingerprint in order to
clear the correct caches and update persistent state for system changes.
On devices that no longer receive updates from the OEM, the build
fingerprint never changes and Android doesn't account for updates
correctly, which causes issues when updating without wiping data.
Only spoofing the fingerprint for Google Play Services fixes this issue.

Corresponding vendor commit:
"Only use stock build fingerprint for Google Play Services"

NB: This code is under the gmscompat package, but it does not depend on
any code from gmscompat.

Change-Id: I26a2498eb2e2163933303b03f6d516e5fb30fe51

### DIFF
--- a/core/java/android/app/Instrumentation.java
+++ b/core/java/android/app/Instrumentation.java
@@ -57,6 +57,7 @@ import android.view.WindowManagerGlobal;
 
 import com.android.internal.content.ReferrerIntent;
 
+import com.android.internal.gmscompat.AttestationHooks;
 import com.android.internal.gmscompat.GmsHooks;
 
 import java.io.File;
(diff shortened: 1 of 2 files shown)


flawedworld:
It doesn't seem that this will be feasible.

PR #170, opened 2022-03-28
gmscompat: improve Dynamite module support 

muhomorr:
- support `DelegateLastClassLoader`s that don't define `librarySearchPath`
(eg DynamiteLoader.apk) by checking `dexPath` too
- make the class loader path check faster now that there's two of them
- enable Dynamite modules in Play Store processes
- use the "/proc/self/fd" fd passing approach for opening the dex files
- don't track the module loading state, cache file descriptor lookups
instead. Previous approach stopped working when module loading sequence
changed (extra `File.lastModified()` check was added)
- use a correct zipFileSeparator ("!/", not "!")
- support `ApkAsset`s that use an `AssetsProvider`
- use "/proc/self/fd" path for `File.lastModified()` hook
instead of an extra IPC to Play services

See the commit messages for more info.

Needs to be applied together with
https://github.com/GrapheneOS/platform_art/pull/1
https://github.com/GrapheneOS/platform_libcore/pull/6
https://github.com/GrapheneOS/platform_packages_apps_GmsCompat/pull/11


ghost (changes requested):
```genfscon
```

PR #190, opened 2022-04-21
warn before uninstalling GMS if it's present in other user profiles

muhomorr:
Depends on https://github.com/GrapheneOS/platform_packages_apps_Settings/pull/91

Helps with https://github.com/GrapheneOS/os-issue-tracker/issues/919


muhomorr:
Fixed in AOSP 13.

PR #192, opened 2022-04-21
Ignore allowbackup on all API versions

chriswoope:
This changes the backup eligibility rules so that the new feature that ignores allowbackup for D2D backups is extended to apply to applications of all API versions and not just those targeting Android 12 or later.

Many applications seem to set allowbackup="false" for no reason, and this allows the user to properly backup all application data with SeedVault and not lose data if the device is lost or destroyed.

### DIFF
--- a/services/backup/java/com/android/server/backup/utils/BackupEligibilityRules.java
+++ b/services/backup/java/com/android/server/backup/utils/BackupEligibilityRules.java
@@ -78,7 +78,7 @@ public class BackupEligibilityRules {
      */
     @ChangeId
     @Overridable
-    @EnabledSince(targetSdkVersion = Build.VERSION_CODES.S)
+    @EnabledSince(targetSdkVersion = Build.VERSION_CODES.BASE)
     static final long IGNORE_ALLOW_BACKUP_IN_D2D = 183147249L;
 
     public static BackupEligibilityRules forBackup(PackageManager packageManager,


thestinger:
This isn't a safe change. It's tied to API level because it changes the semantics of an API in a backwards incompatible way which can easily break applications or cause security issues. It's not going to be long until API 31 is required for all applications on the Play Store. The correct solution is being patient and waiting for them all to be updated.

thestinger:
The way you're changing this also doesn't do what you're intending. You're only changing the API level where this constant is considered to be available. You're not actually changing the API level where this applies in the code implementing it.

chriswoope:
Unfortunately there is no guarantee that apps receive updates or that if they do they will be updated to new API SDK levels (that's only a Play Store requirement, and there is no need to distribute on Play Store); furthermore, people need to backup their data now, not when the application developer decides out of their goodwill to gracefully allow them the privilege to perform this fundamental action, so waiting is not really a good solution.

It would be better though to put this change behind a setting for maximum compatibility with existing workflows.

As for the implementation, what would be the proper way to change it? There are no other mentions of IGNORE_ALLOW_BACKUP_IN_D2D in the Android source code outside tests other than the code in BackupEligibilityRules that applies it which includes no other API version level, so unless the value of the field is meaningful other than a generated ID, I'm not seeing where the API version level to check against the application could be other that the code I changed, unless the change IDs are baked into the APK at build time, and it automatically finds and bakes all the applicable ones? (although I have not tested the change and you know the code way better than me, so I guess I'm missing something?).

chriswoope:
Maybe a better implementation could be to change the OS so that all apps are assumed to be debuggable (a very good change on its own since it removes another user-hostile restriction) and then have the user go into the App Compatibility Changes developer settings and turn on IGNORE_ALLOW_BACKUP_IN_D2D manually for the apps; I'm not sure if this survives app updates though.

thestinger:
You're making proposals compromising the security model of the OS. It's not going to be implemented.

thestinger:
> Unfortunately there is no guarantee that apps receive updates or that if they do they will be updated to new API SDK levels

There are hard and soft minimum API levels in the OS already and those are going to be increased over time. Apps unable to run within the strictest modern sandbox aren't going to be supported indefinitely. Many legacy apps only have 32-bit native libraries and will be dropped when 32-bit support goes away. Long-term support for completely unmaintained apps belongs in a virtual machine running a legacy OS version. Unmaintained apps often gradually break over time since a lot of apps depend on implementation details not explicitly defined in the APIs. Not every single behavior change is gated for new API levels only, only changes that are expected to cause any significant backwards compatibility issues (like this one) or are found to cause them in practice during months of beta testing.

API 31 was released in October last year, and will end up becoming required for both new apps and updates to existing updates on the Play Store near the end of the year. The Play Store is also going to be hiding apps falling 2 years behind the current API level for new users. It's expected that every app on the Play Store at least has minimal maintenance.

>  that's only a Play Store requirement, and there is no need to distribute on Play Store

Lower privacy and security standards for apps installed from other sources can be addressed through enforcing the same requirements in the OS, and that also avoids needing to trust that the Play Store properly enforces the requirements they set.

> It would be better though to put this change behind a setting for maximum compatibility with existing workflows.

That doesn't really address the compatibility issues especially since users won't understand how badly things can break. It doesn't address the security issues either.

thestinger:
> 
Maybe a better implementation could be to change the OS so that all apps are assumed to be debuggable (a very good change on its own since it removes another user-hostile restriction) and then have the user go into the App Compatibility Changes developer settings and turn on IGNORE_ALLOW_BACKUP_IN_D2D manually for the apps; I'm not sure if this survives app updates though.

If you want to run modified versions of apps, you can do that without OS changes. You're free to mark an app debuggable and sign it with your own signing key. GrapheneOS isn't going to compromise OS security and break multiple aspects of the security model to support something you can do without OS support.

thestinger:
> Maybe a better implementation could be to change the OS so that all apps are assumed to be debuggable (a very good change on its own since it removes another user-hostile restriction) and then have the user go into the App Compatibility Changes developer settings and turn on IGNORE_ALLOW_BACKUP_IN_D2D manually for the apps; I'm not sure if this survives app updates though.

If you want to run modified versions of apps, you can do that without OS changes. You're free to mark an app debuggable and sign it with your own signing key. GrapheneOS isn't going to compromise OS security and break multiple aspects of the security model to support something you can do without OS support.

thestinger:
> Maybe a better implementation could be to change the OS so that all apps are assumed to be debuggable (a very good change on its own since it removes another user-hostile restriction) and then have the user go into the App Compatibility Changes developer settings and turn on IGNORE_ALLOW_BACKUP_IN_D2D manually for the apps; I'm not sure if this survives app updates though.

If you want to run modified versions of apps, you can do that without OS changes. You're free to mark an app debuggable and sign it with your own signing key. You can similarly change properties like this one in the manifest.

GrapheneOS isn't going to compromise OS security and break multiple aspects of the security model to support something you can do without OS support.

chriswoope:
The problem is that the Android security models appears to serve two purposes: one is to prevent a malicious attacker from having the device work in their interest instead of the user's interest, and the second is to prevent the user from having the device work the way they want instead of in Google's, the phone manufacturers' or the app developers' interest.

The latter presumably happens because they want to increase their profits (or indirectly respond to pressure by those seeking that) by making it hard for the user to do things like fake IAP purchases or extract DRMed media files or making it hard to cheat in games, and so on. This might also be in the interest of users as a whole (maybe there would be less incentive to develop commercial software, which might be useful), but is definitely not in the interest of the single user, especially when it prevents him from doing things that aren't even against other people's interest, like backing up their down data.

In the case of application backups, it appears that Google's main goal in adding the ability for the app developer to disable them might have been to prevent users from examining app data and possibly to mitigate some technical limitations, and it might have been then co-opted by developers to prevent the Google backups from reading their data and giving it to Google. The latter is not an issue in GrapheneOS since Seedvault is used, and the former is not a problem since it's not in the user's interest to prevent themselves from examining app data.

There is a danger though: any feature that makes it possible to expand the user's ability to act on the device also allows a successful exploiters to leverage it for their interests (e.g. someone exploiting SeedVault can then read more app data), and also allows them to social engineer the user to use it for their interests (e.g. they can try to convince the user to do a SeedVault backup pointing to their NextCloud server and give them the 12 words code, which would then give them more app data with the change). But this should not be avoided absolutely because otherwise the resulting decision would be to not have any backup system at all (and in general drop a lot of features), and while there's a legitimate tradeoff there I feel like a choice should not be made that prevents basic functionality like backing up all data on the device.

So in this case, to summarize, it seems like allowbackup is mostly an anti-user feature and is also used to protect against issues and threats not present on GrapheneOS, and should thus be removed.

Resigning apks is another possible approach, but is far more complex to implement and even more extreme in the security tradeoff than this patch. In particular, to support adding apps and keeping automatic updates working easily it's necessary to do the resigning on the device (and have F-Droid and Aurora Store automatically resign), and to keep application compatibility the OS must be patched to support signature spoofing for whitelisted keys so that applications cannot tell that they or others have been resigned; this effectively results in a much more powerful mechanism that the proposed change and requires more invasive OS changes for the signature spoofing.

chriswoope:
The apk resigning vs OS modification decision is also similar to what has been done with Sandboxed Play Services, which has been implemented with somewhat invasive modifications of GrapheneOS libraries instead of patching the DEX code and resigning the play services APKs and adding signature spoofing to the OS, presumably because it's simpler.

thestinger:
> The problem is that the Android security models appears to serve two purposes: one is to prevent a malicious attacker from having the device work in their interest instead of the user's interest, and the second is to prevent the user from having the device work the way they want instead of in Google's, the phone manufacturers' or the app developers' interest.

The backup file exclusive lists and release builds not being debuggable act as valuable security features protecting users. Bypassing them violates the security model of the OS in a way that negatively impacts end users. GrapheneOS isn't going to do be doing that.

Your change is not doing anything for apps targeting API 31 or later. API 31 or later will be required by the end of this year for all new apps and app updates on the Play Store. Apps which haven't moved to the new API level after that point (meaning they cannot ship any more app updates) will be unlisted from the Play Store after around another year to a year and a half. As was already stated above, you simply need to be patient and API 31 or later will already be the norm. The soft enforced minimum API level requirement will be raised in GrapheneOS to match the Play Store requirement at some point. The hard enforced minimum API level requirement will eventually started to be raised too.

> The latter presumably happens because they want to increase their profits (or indirectly respond to pressure by those seeking that) by making it hard for the user to do things like fake IAP purchases or extract DRMed media files or making it hard to cheat in games, and so on. This might also be in the interest of users as a whole (maybe there would be less incentive to develop commercial software, which might be useful), but is definitely not in the interest of the single user, especially when it prevents him from doing things that aren't even against other people's interest, like backing up their down data.

Not clear how this has anything to do with this pull request or these topics.

> In the case of application backups, it appears that Google's main goal in adding the ability for the app developer to disable them might have been to prevent users from examining app data and possibly to mitigate some technical limitations, and it might have been then co-opted by developers to prevent the Google backups from reading their data and giving it to Google. The latter is not an issue in GrapheneOS since Seedvault is used, and the former is not a problem since it's not in the user's interest to prevent themselves from examining app data.

There's already a distinction between cloud and device-to-device backups. Again, you need to be patient and wait another half year for apps to finish being updated instead of trying to rush this new feature in a way that violates the security model. That's never going to fly for GrapheneOS, sorry.

> So in this case, to summarize, it seems like allowbackup is mostly an anti-user feature and is also used to protect against issues and threats not present on GrapheneOS, and should thus be removed.

allowBackup is a legacy toggle which only disables cloud backups. It's not an anti-user feature. What matters is how it works now, not how it used to work before current API levels. Wait 6 months, and apps not updated to API level 31 will be an extreme minority, since the vast majority of apps have releases on the Play Store and have to keep up with the requirements. If the app developer is shipping variants with a lower target API outside the Play Store, that's highly suspicious and almost certainly being done to bypass backwards incompatible privacy/security restrictions. Generally apps distributed outside the Play Store have the same target API level. It's a strong sign of malicious behavior if they use a lower target API outside the Play Store.

I don't think it makes any sense to have lower privacy and security standards for apps outside the Play Store in general, and GrapheneOS is going to be continuing to raise the soft enforced target API level to more closely match the Play Store along with making it a bigger barrier than a first launch warning.

> Resigning apks is another possible approach, but is far more complex to implement and even more extreme in the security tradeoff than this patch. In particular, to support adding apps and keeping automatic updates working easily it's necessary to do the resigning on the device (and have F-Droid and Aurora Store automatically resign), and to keep application compatibility the OS must be patched to support signature spoofing for whitelisted keys so that applications cannot tell that they or others have been resigned; this effectively results in a much more powerful mechanism that the proposed change and requires more invasive OS changes for the signature spoofing.

That's not at all accurate.

ghost:
> The apk resigning vs OS modification decision is also similar to what has been done with Sandboxed Play Services, which has been implemented with somewhat invasive modifications of GrapheneOS libraries instead of patching the DEX code and resigning the play services APKs and adding signature spoofing to the OS, presumably because it's simpler.

It's not "invasive". It has zero privileged or special integration. All we do is use the existing app APIs that Play services doesn't know how to use and make Play services use the existing APIs that **any** other app can use.

Signature spoofing is substantially harmful and breaks the security model. It allows apps to intercept data from other apps that are supposed to be signature-bound. We're not going to patch or resign anything. It would break a ton of things and would be far more complicated to maintain each Play services release.

You seem to be claiming that security is "anti-user" here and trying to request extreme downgrades in OS security here. They are not getting merged or accepted. You can make your own GrapheneOS builds with all the security features removed and then have what is just a far-worse stock OS. We're not going to take the easy way out and integrate insecure and privacy-theatre libraries and services.

thestinger:
> and to keep application compatibility the OS must be patched to support signature spoofing for whitelisted keys so that applications cannot tell that they or others have been resigned

You can modify this along with the rest of what you're modifying. Hardware-based attestation can be used to detect what's happening just as it can be used to detect an OS that's not upholding the security model. As is GrapheneOS upholds the whole security model and can reasonably ask apps to whitelist it. This guide would need to be removed if we implemented your proposals and we'd need to give up on it.

https://grapheneos.org/articles/attestation-compatibility-guide

> The apk resigning vs OS modification decision is also similar to what has been done with Sandboxed Play Services, which has been implemented with somewhat invasive modifications of GrapheneOS libraries instead of patching the DEX code and resigning the play services APKs and adding signature spoofing to the OS, presumably because it's simpler.

Sandboxed Google Play doesn't bypass security checks or the security model and there's no way to implement it without either OS changes or official Google Play support for sandboxed apps. The latter would not provide location API re-routing unless the client-side Play libraries implemented it.

chriswoope:
> Not clear how this has anything to do with this pull request or these topics.

It's there to point out that the Android security model should NOT be followed blindly, since it's partially designed to ensure that someone else's interests are safeguarded at the expense of the user.

> Signature spoofing is substantially harmful and breaks the security model. It allows apps to intercept data from other apps that are supposed to be signature-bound

Obviously signature spoofing would only be allowed for apps signed by a key set by the user, so it doesn't really break the "against attackers" security model assuming the device is not exploited (it does break the "against the user" security model, since it lets the user examine all data on their own device, against the interest of third-parties).

> As is GrapheneOS upholds the whole security model and can reasonably ask apps to whitelist it. This guide would need to be removed if we implemented your proposals and we'd need to give up on it. https://grapheneos.org/articles/attestation-compatibility-guide

If you really want to retain the ability to let the user attest to third-parties that their OS is ensuring that their interests override the user's interest, you could make two variants of GrapheneOS signed with different keys, where only one acts in the user's interest and the hypothetical third-party could then use attestation to tell which version the user is running (it doesn't need to be two different builds, they can just be the same build signed with different keys and with logic enabled depending on the AVB key hash of the device).

Of course having remote attestation usable by third parties in general is itself an anti-user feature (it lets the third-party ensure that the device is not working in the user's interest), but since Google provides it, third-parties can already require a Google-attested device, so it's not a big issue if GrapheneOS also allows it.

PR #193, opened 2022-04-22
Decrease minimum color temperature for Night Light

psocketer:
Decrease _minimum_ color temperature to 500K from stock value of 2596K, to allow for further decrease in blue light. Followed reference [spec](https://source.android.com/devices/tech/display/night-light) for Night Light. Tested on Pixel 4XL.


ghost:
It isn't a problem since we can just change the commit message ourselves, but in the future please make sure your commit messages are detailed and useful.

"Decrease minimum colour temperature for Night Light"

psocketer:
> It isn't a problem since we can just change the commit message ourselves, but in the future please make sure your commit messages are detailed and useful.
> 
> "Decrease minimum colour temperature for Night Light"

Certainly, my apologies, I didn't intend to do it that way. Thanks for bringing that to my attention.

flawedworld:
Where did you get this value from?

flawedworld:
This value this is set to should be pulled from the stock OS on a per device basis if that's what this is aiming to do.

psocketer:
I wanted to override this stock value to provide a level of red shifting greater than that from stock. The current value is far too blue and is practically useless.

psocketer:
Let's say I did find a way to pull it from stock on a per-device basis, and the values were still "too high" to be useful. Would going below stock, but based on initial per-device values, be acceptable? If so, I'll pursue doing it.

psocketer:
I checked the device framework overlays for Pixel 6 to Pixel 3 and I don't see any of them modifying this stock value for minimum color temperature.

flawedworld:
There might not be an overlay but you should check ```/system/framework/framework-res.apk``` to see if it differs for this option.

psocketer:
Decompiled the stock `/system/framework/framework-res.apk` and found no settings related to this option. Checked the behavior on stock ROM and it appears to have the same minimum color temp. Seems like stock is using the default values. Based on [this issue](https://github.com/GrapheneOS/os-issue-tracker/issues/695), this might not be something that is within the scope of GrapheneOS, but it would be extremely helpful to have an actual amber screen and not a yellow one; getting something like this included upstream is also very unlikely. 

Google seems to suggest it is within the purview of device developers to adopt these settings, so wondering why this cannot be modified. Is the concern about hardware damage?

ghost:
I think the number just has to be made with some reason behind it. 2596 -> 500 seems very arbitrary to me and 2596 looks like it was picked with a reason.

psocketer:
I'm not sure it's possible to figure out what the author of that default minimum value had in mind. I did try searching the Android issue database and the codebase for references to, or comments about, this value, and found nothing. 

That said, I can at least come up with a plausible explanation (or at least interesting "coincidences"). The color temperature values of 2600K to 4100K (which are approximately what given for the min/max) are roughly those ranges between standard incandescent bulbs and fluorescent bulbs. If you account for the fact that users will be using a slider, and want to make sure they have access to the full range, perhaps they padded their value ranges toward the red end. 

Further, if you look at the plot of the color intensity functions mentioned in the Night Light [spec](https://source.android.com/devices/tech/display/night-light), over the range of values between 2596K and 4082K, you see that the range of intensities (compared to red at a value of 1.0) for blue are 0.459 to 0.72 and those for green are 0.723 to 0.855, as if they chose these values so that the range of blue intensity never overlaps with the green intensity range. 

Another reason to have chosen a higher minimum value could also have been to mitigate complaints from users accidentally setting their intensity values (along with brightness settings) too low so as to be difficult to read the screen. Or, there could be some hardware reason, though not all Pixels have the same screen manufacturers, gamuts, etc. but yet use the same default values.

Regardless, I think that there are a significant number of users who would prefer to be able to set their Night Light intensity to be more amber. It would not take away any feature, but rather just allow users to set it lower, unless there is a hardware reason for not allowing a lower value.

I agree that 500K is completely arbitrary. I chose it as a random, whole number that gave the effect I wanted. It is also the value currently being used by another pixel-based ROM that I switched from. If I were to suggest a less arbitrary value, I would choose 686K, as this is where the blue intensity function hits zero. And, since the purpose is to mainly remove the blue light, this would allow users to choose to remove as much blue light as possible.

ghost:
I validated 686K and works on my 6 Pro. Would appreciate testing on other devices after this new release.

https://github.com/GrapheneOS/platform_frameworks_base/commit/1ab1240be9ed74bd19fc70ebb5e4ff7ccea4f580

PR #199, opened 2022-05-11
don't crash apps that try to access the "NETWORK" location provider; + hidden Play Store option to allow GMS updates

muhomorr:
Reroute them to the "PASSIVE" provider. Note that the "NETWORK" provider
is still not advertised as being present (it will not be returned
by `getAllProviders()`, `getBestProvider()`, `hasProvider()`).

There's a relevant system feature (`android.hardware.location.network`)
that is still advertised as being present to improve app compatibility.
Removing it would not make this change redundant, because there are apps
that try to use the "NETWORK" provider without declaring the need for
this feature or checking its presence at runtime.

### DIFF
--- a/location/java/android/location/LocationManager.java
+++ b/location/java/android/location/LocationManager.java
@@ -1725,6 +1725,13 @@ public class LocationManager {
     public boolean hasProvider(@NonNull String provider) {
         Preconditions.checkArgument(provider != null, "invalid null provider");
 
+        if (NETWORK_PROVIDER.equals(provider)) {
+            if (!getAllProviders().contains(NETWORK_PROVIDER)) {
+                android.util.Log.d("LocationManager", "returning false from hasProvider(NETWORK_PROVIDER)");
+                return false;
+            }
+        }
+
         try {
             return mService.hasProvider(provider);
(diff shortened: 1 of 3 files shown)


location/java/android/location/LocationManager.java
```diff
@@ -1725,6 +1725,11 @@ public void removeUpdates(@NonNull PendingIntent pendingIntent) {
     public boolean hasProvider(@NonNull String provider) {
         Preconditions.checkArgument(provider != null, "invalid null provider");
 
+        if (NETWORK_PROVIDER.equals(provider)) {
+            android.util.Log.d("LocationManager", "returning false from hasProvider(NETWORK_PROVIDER)");
+            return false;
```
flawedworld: Whilst not essential for official GrapheneOS releases, we should change this to check if there are any network location providers within the OS before we return false at some point.

muhomorr: There can be only one "network" location provider (`NETWORK_PROVIDER` refers to the "network" string constant).
If it's present, this commit should be dropped. However, it's simple to make it so that this commit doesn't break the "real" network location provider, I'll update it now.

PR #201, opened 2022-05-18
disable creation of shared relro files

flawedworld:


thestinger:
Might not make sense anymore now that exec spawning is optional. The attack surface from this is quite low since it's not actually executable data.

PR #206, opened 2022-06-06
add com.android.location.fused to indicator exemptions

flawedworld:
Indicator misleadingly goes off sometimes


thestinger:
Saw this after making the change myself.

PR #215, opened 2022-06-25
Replace deprecated NetworkInfo in Wi-Fi timeout code

lukashoracek:
https://github.com/GrapheneOS/os-issue-tracker/issues/1205

Not tested.


lukashoracek:
This needs some testing especially with always-on VPN & block connections without VPN

PR #232, opened 2022-08-01
Immediately put into effect always on vpn on first connection

quh4gko8:
Don't require reboot or settings re-set for always on and lockdown to take effect on first vpn connection.

Requires https://github.com/GrapheneOS/platform_packages_modules_Connectivity/pull/10


quh4gko8:
Closing in favor of #233

PR #233, opened 2022-08-01
Immediately put into effect always on vpn on first connection

quh4gko8:


quh4gko8:
Don't require reboot or settings re-set for always on and lockdown to take effect on first vpn connection. The requirement for reboot, re-set at settings has been caused by a permission not granted or declared by VpnDialogs, the app handles prompt on first vpn action, which caused the unexpected behavior. Prevent this by checking the permission of local process instead.

PR #255, opened 2022-08-23
Change Strings to Focus on Ended Session Rather than Device Restart

inferenceus:
Make user profile ended session clearer due to restart not necessarily the reason for an ended session. I believe this change would make it clearer to users that the session has ended by any means, instead of only by device restart.


thestinger:
Going to be deferring this and other changes until post-13 migration (i.e. after we ship 13 in Stable channel).

I agree with the concept of this change but I'm not sure it's as clear as it needs to be since people might not interpret session ended correctly.

inferenceus:
Thanks for considering the change. Perhaps changing the strings to something more like "logged out" may be preferable.

PR #285, opened 2022-12-02
allow configuring notification for cell broadcast

empratyush:
closes [#1759](https://github.com/GrapheneOS/os-issue-tracker/issues/1759)


thestinger:
Had to revert this since I'm not clear on how the configuration is going to work when handling the default special behavior.

PR #286, opened 2022-12-03
always auto-grant permissions that have "normal" protectionLevel

muhomorr:


thestinger:
Merged the auto-grant normal permissions commit, other commit I'd rather have replaced with a more generic approach adding a generic notice about shared uid to install request UI.

PR #289, opened 2022-12-10
Enable clock customization

flawedworld:


flawedworld:
AOSP does not have the right infra for this yet. Revisit.

PR #291, opened 2022-12-13
32 bit warning for 32 bit apps

flawedworld:


flawedworld:
Android 14 will have this.

PR #293, opened 2022-12-14
Revert "add 2 button navigation overlay"

### LINKED ISSUE #1541: Rotate Suggestion not working in 2 button navigation
Gesture nav sticks to the bottom of the screen in landscape instead of moving to the side like with 3 and 2 button nav, so I am forced to use 2 button so that I am able to type effectively in landscape... In switching I discovered the screen rotation hint is not present on 2 button but is on 3 and gesture nav.

Issue and fix from another project: https://github.com/PixelExperience/android-issues/issues/2389

Pixel 6 Pro, GOS latest stable (TP1A.221005.002.2022100300)

neelchauhan:
This reverts commit d2521b78e39acf4553e9df5aa094671f863cc6fb.

Closes https://github.com/GrapheneOS/os-issue-tracker/issues/1541


x86pup:
The commit for adding the 2 button overlay was dropped instead. https://github.com/GrapheneOS/platform_frameworks_base/blob/13/packages/overlays/Android.mk#L32-L33

PR #296, opened 2022-12-23
Set config_clockFontFamily to empty string

maade93791:
This will match pre QPR1 clock font.

Change-Id: Idef5fb093d81f49e40f2fcf4df69f834804916ae


thestinger:
Cherry-picking this.

thestinger:
Actually, nevermind. I think we do want it monospace. We just want a better monospace font. We should see what the stock OS is doing.

PR #302, opened 2023-01-06
Better defaults

x86pup:
Makes the phone more quiet by default, AOSP sounds suck anyways

Turns off location by default because Wi-Fi and Bluetooth are already turned off by default. The user can enable it just fine in SetupWizard like normal.

Solves https://github.com/GrapheneOS/os-issue-tracker/issues/1133 by enabling auto brightness and auto-rotation (unsure if wanted but can remove)


thestinger:
Not sure about the auto-rotation change since it's much less useful now that there's a rotation button appearing when you rotate the phone rather than needing to open quick settings.

Need to think more about the sound changes.

thestinger:
Cherry-picked the auto brightness change for now.

matchboxbananasynergy:
Would it make sense/be wanted to switch `<bool name="def_airplane_mode_on">false</bool>` to "true" here?

Many people use GrapheneOS as Wi-Fi only, and defaulting to airplane mode makes it so the device doesn't connect to cell towers by default.

It might make sense to not do this until the new SetupWizard (when that comes), at which point an option can be added to disable airplane mode during setup for people who want to.

thestinger:
People will be confused by it unless it's included in the setup wizard.

PR #311, opened 2023-01-21
Ignore setting filterCallingUid now instant apps are formally unsupported

quh4gko8:
This parameter is introduced to [workaround resolving with instant intents](https://github.com/GrapheneOS/platform_frameworks_base/commit/31426b26), but this is now formally unsupported. Furthermore, when resolving activity, [apps filter is ignored](https://cs.android.com/android/platform/superproject/+/android-13.0.0_r18:frameworks/base/services/core/java/com/android/server/pm/ComputerEngine.java;l=1371;drc=72a1e39e4baa3068368c5fd2f4c8b1dc1aa08ffd), which is set on [resolving intent](https://cs.android.com/android/platform/superproject/+/android-13.0.0_r18:out/soong/.intermediates/frameworks/base/services/core/services.core.unboosted/android_common/xref32/srcjars.xref/frameworks/base/services/core/java/com/android/server/wm/ActivityTaskSupervisor.java;l=748;bpv=1;bpt=1)


thestinger:
Can this throw a RuntimeException instead?

quh4gko8:
It will crash calls from `Activity#startActivityAsCaller`, such as those being called by ResolverActivity or ChooserActivity, and avoid properly resolving, hence ignoring it is done for compatibility

quh4gko8:
Alternatively, every present usage (just one at `ActivityTaskManagerService#startActivityAsCaller`) of `ActivityStarter.Request#setFilterCallingUid`  can be removed and throw the said exception to be notified for its future usages

thestinger:
cc @muhomorr

PR #318, opened 2023-01-31
Always treat instant apps as disabled for all users

quh4gko8:


quh4gko8:
https://github.com/GrapheneOS/platform_frameworks_base/pull/311 will be included here as part of  further disabling instant apps suppirt

PR #320, opened 2023-02-01
Disable instant apps feature and settings, with adding users at lock screen every boot

quh4gko8:
Merge with https://github.com/GrapheneOS/platform_packages_apps_Settings/pull/140 for completeness, that is, to hide the now no-op Settings that toggle the moved settings

### DIFF
--- a/packages/SettingsProvider/src/com/android/providers/settings/SettingsProvider.java
+++ b/packages/SettingsProvider/src/com/android/providers/settings/SettingsProvider.java
@@ -2919,6 +2919,7 @@ public class SettingsProvider extends ContentProvider {
             // Upgrade the settings to the latest version.
             UpgradeController upgrader = new UpgradeController(userId);
             upgrader.upgradeIfNeededLocked();
+            SettingsProviderHooks.onSettingsStateInit(SettingsRegistry.this, userId);
             return true;
         }
 
--- a/packages/SettingsProvider/src/com/android/providers/settings/SettingsProviderHooks.java
+++ b/packages/SettingsProvider/src/com/android/providers/settings/SettingsProviderHooks.java
@@ -9,6 +9,7 @@ class SettingsProviderHooks {
         if (userId == UserHandle.USER_SYSTEM) {
             SettingsState globalSettings = registry.getSettingsLocked(SettingsProvider.SETTINGS_TYPE_GLOBAL, userId);
             insertSetting(globalSettings, Settings.Global.ADD_USERS_WHEN_LOCKED, "0" /* disabled value */);
+            insertSetting(globalSettings, Settings.Global.ENABLE_EPHEMERAL_FEATURE, "0" /* disabled value */);
         }
         SettingsState secureSettings = registry.getSettingsLocked(SettingsProvider.SETTINGS_TYPE_SECURE, userId);
     }
--- a/packages/SettingsProvider/src/com/android/providers/settings/SettingsProviderHooks.java
+++ b/packages/SettingsProvider/src/com/android/providers/settings/SettingsProviderHooks.java
@@ -1,0 +1,22 @@
+package com.android.providers.settings;
+
+import android.os.UserHandle;
+import android.provider.Settings;
+
+class SettingsProviderHooks {
+
+    static void onSettingsStateInit(final SettingsProvider.SettingsRegistry registry, final int userId) {
+        if (userId == UserHandle.USER_SYSTEM) {
+            SettingsState globalSettings = registry.getSettingsLocked(SettingsProvider.SETTINGS_TYPE_GLOBAL, userId);
+        }
+        SettingsState secureSettings = registry.getSettingsLocked(SettingsProvider.SETTINGS_TYPE_SECURE, userId);
+    }
+
+    /**
+     * Unlike UpgradeController#upgradeIfNeededLocked settings migration, this runs every time a user is initialized.
+     * Insert or modify setting upon SettingState initialization for any user, or in case of system user, upon boot.
+     */
+    private static void insertSetting(SettingsState state, String key, String value) {
+        state.insertSettingLocked(key, value, null /* tag */, false /* makeDefault */, SettingsState.SYSTEM_PACKAGE_NAME);
+    }
+}
--- a/packages/SettingsProvider/src/com/android/providers/settings/SettingsProviderHooks.java
+++ b/packages/SettingsProvider/src/com/android/providers/settings/SettingsProviderHooks.java
@@ -8,6 +8,7 @@ class SettingsProviderHooks {
     static void onSettingsStateInit(final SettingsProvider.SettingsRegistry registry, final int userId) {
         if (userId == UserHandle.USER_SYSTEM) {
             SettingsState globalSettings = registry.getSettingsLocked(SettingsProvider.SETTINGS_TYPE_GLOBAL, userId);
+            insertSetting(globalSettings, Settings.Global.ADD_USERS_WHEN_LOCKED, "0" /* disabled value */);
         }
         SettingsState secureSettings = registry.getSettingsLocked(SettingsProvider.SETTINGS_TYPE_SECURE, userId);
     }


packages/SystemUI/src/com/android/systemui/statusbar/policy/UserSwitcherControllerOldImpl.java
```diff
@@ -257,11 +257,6 @@ public UserSwitcherControllerOldImpl(
             @Override
             public void onChange(boolean selfChange) {
                 mSimpleUserSwitcher = shouldUseSimpleUserSwitcher();
-                mAddUsersFromLockScreen.setValue(
```
quh4gko8: This removal makes the `mAddUsersFromLockScreen` remain at default false value as defined above

quh4gko8: https://cs.android.com/android/platform/superproject/+/android-13.0.0_r18:frameworks/base/packages/SystemUI/src/com/android/systemui/statusbar/policy/UserSwitcherControllerOldImpl.java;l=155-156

quh4gko8: This resolves https://github.com/GrapheneOS/os-issue-tracker/issues/1927

services/core/java/com/android/server/pm/PackageManagerService.java
```diff
@@ -4010,13 +4010,8 @@ public void systemReady() {
         ContentObserver co = new ContentObserver(mHandler) {
```
muhomorr: I think removing this ContentObserver entirely and adding
```
                for (int userId : UserManagerService.getInstance().getUserIds()) {
                    mWebInstantAppsDisabled.put(userId, true);
                }
```
in its place would be a better solution.

core/java/android/provider/Settings.java
```diff
@@ -16024,6 +16033,16 @@ public static final class Global extends NameValueTable {
...
         };
 
+        /**
+         * Global settings that are moved to non-persisted state.
+         *
+         * @hide
+         */
+        public static final String[] MOVED_TO_TRANSIENT_SETTINGS = {
+                ADD_USERS_WHEN_LOCKED,
+                ENABLE_EPHEMERAL_FEATURE,
```
muhomorr: I think a better solution would be to remove definitions of all these keys.
It'll ensure that all places that read them are dealt with.

packages/SettingsProvider/src/com/android/providers/settings/SettingsState.java
```diff
@@ -1153,6 +1153,9 @@ private void parseSettingsLocked(TypedXmlPullParser parser)
                     fromSystem = parser.getAttributeBoolean(null, ATTR_DEFAULT_SYS_SET, false);
                     tag = getValueAttribute(parser, ATTR_TAG, ATTR_TAG_BASE64);
                 }
+                if (isSettingBlockedFromReadAndWrite(name, mKey)) {
```
muhomorr: This can be moved up to line 1144 (after `name` is set) to avoid wasting time obtaining other attributes.

packages/SettingsProvider/src/com/android/providers/settings/SettingsState.java
```diff
@@ -1285,6 +1314,9 @@ public Setting(String name, String value, String defaultValue,
         private void init(String name, String value, String tag, String defaultValue,
                 String packageName, boolean fromSystem, String id,
                 boolean isValuePreservedInRestore) {
+            if (isSettingBlockedFromReadAndWrite(name, getKey())) {
```
muhomorr: Which operation does this block?

quh4gko8: Moving checks on places where `mSettings.put()` is called

quh4gko8:
for simplicity, it will be reset every boot as disabled instead

packages/SettingsProvider/src/com/android/providers/settings/SettingsProvider.java
```diff
@@ -2919,6 +2919,19 @@ public boolean ensureSettingsForUserLocked(int userId) {
...
+            if (userId == UserHandle.USER_SYSTEM) {
+                SettingsState globalSetting = mSettingsStates.get(makeKey(SETTINGS_TYPE_GLOBAL, UserHandle.USER_SYSTEM));
+                globalSetting.insertSettingLocked(Settings.Global.ADD_USERS_WHEN_LOCKED,
+                        "0" /* disabled */, null /* tag */, false /* makeDefault */,
+                        SettingsState.SYSTEM_PACKAGE_NAME);
+                globalSetting.insertSettingLocked(Settings.Global.ENABLE_EPHEMERAL_FEATURE,
+                        "0" /* disabled */, null /* tag */, false /* makeDefault */,
+                        SettingsState.SYSTEM_PACKAGE_NAME);
+            }
+            SettingsState secureSetting = mSettingsStates.get(secureKey);
```
muhomorr: `secureSettings` are already retrieved above

packages/SettingsProvider/src/com/android/providers/settings/SettingsProvider.java
```diff
@@ -2919,6 +2919,19 @@ public boolean ensureSettingsForUserLocked(int userId) {
             // Upgrade the settings to the latest version.
             UpgradeController upgrader = new UpgradeController(userId);
             upgrader.upgradeIfNeededLocked();
+            if (userId == UserHandle.USER_SYSTEM) {
+                SettingsState globalSetting = mSettingsStates.get(makeKey(SETTINGS_TYPE_GLOBAL, UserHandle.USER_SYSTEM));
+                globalSetting.insertSettingLocked(Settings.Global.ADD_USERS_WHEN_LOCKED,
```
muhomorr: `globalSettings` would be a more consistent name

quh4gko8: I'll just reassign `secureSettings`

quh4gko8: to be able to use same consistent name

packages/SettingsProvider/src/com/android/providers/settings/SettingsProvider.java
quh4gko8: okay, can use that instead

packages/SettingsProvider/src/com/android/providers/settings/SettingsProvider.java
```diff
@@ -2919,6 +2919,7 @@ public boolean ensureSettingsForUserLocked(int userId) {
             // Upgrade the settings to the latest version.
             UpgradeController upgrader = new UpgradeController(userId);
             upgrader.upgradeIfNeededLocked();
+            SettingsProviderHooks.insertSettingsOnBoot(SettingsRegistry.this, userId);
```
muhomorr: This method is also called when user is added, correct? If so, the method name is misleading.

packages/SettingsProvider/src/com/android/providers/settings/SettingsProviderHooks.java
```diff
@@ -24,7 +24,13 @@ class SettingsProviderHooks {
     static void insertSettingsOnBoot(final SettingsProvider.SettingsRegistry registry, final int userId) {
         if (userId == UserHandle.USER_SYSTEM) {
             SettingsState globalSettings = registry.getSettingsLocked(SettingsProvider.SETTINGS_TYPE_GLOBAL, userId);
+            globalSettings.insertSettingLocked(Settings.Global.ADD_USERS_WHEN_LOCKED, "0" /* (disabled) value */,
+                    null /* tag */, false /* makeDefault */, SettingsState.SYSTEM_PACKAGE_NAME);
+            globalSettings.insertSettingLocked(Settings.Global.ENABLE_EPHEMERAL_FEATURE, "0" /* (disabled) value */,
+                    null /* tag */, false /* makeDefault */, SettingsState.SYSTEM_PACKAGE_NAME);
         }
         SettingsState secureSettings = registry.getSettingsLocked(SettingsProvider.SETTINGS_TYPE_SECURE, userId);
+        secureSettings.insertSettingLocked(Settings.Secure.INSTANT_APPS_ENABLED, "0" /* (disabled) value */,
```
muhomorr: This setting is ignored when `ENABLE_EPHEMERAL_FEATURE` is set to 0

packages/SettingsProvider/src/com/android/providers/settings/SettingsProviderHooks.java
```diff
@@ -24,7 +24,13 @@ class SettingsProviderHooks {
     static void insertSettingsOnBoot(final SettingsProvider.SettingsRegistry registry, final int userId) {
         if (userId == UserHandle.USER_SYSTEM) {
             SettingsState globalSettings = registry.getSettingsLocked(SettingsProvider.SETTINGS_TYPE_GLOBAL, userId);
+            globalSettings.insertSettingLocked(Settings.Global.ADD_USERS_WHEN_LOCKED, "0" /* (disabled) value */,
```
muhomorr: To avoid repeating the same argument values, add a wrapper method that takes SettingsState, key and value as parameters.

packages/SettingsProvider/src/com/android/providers/settings/SettingsProviderHooks.java
```diff
@@ -24,7 +24,13 @@ class SettingsProviderHooks {
...
+                    null /* tag */, false /* makeDefault */, SettingsState.SYSTEM_PACKAGE_NAME);
+            globalSettings.insertSettingLocked(Settings.Global.ENABLE_EPHEMERAL_FEATURE, "0" /* (disabled) value */,
```
muhomorr: Split this out into a separate commit

muhomorr (changes requested):

packages/SettingsProvider/src/com/android/providers/settings/SettingsProviderHooks.java
```diff
@@ -0,0 +1,21 @@
...
+
+    static void onSettingsStateInit(final SettingsProvider.SettingsRegistry registry, final int userId) {
+        if (userId == UserHandle.USER_SYSTEM) {
+            SettingsState globalSettings = registry.getSettingsLocked(SettingsProvider.SETTINGS_TYPE_GLOBAL, userId);
+        }
+        SettingsState secureSettings = registry.getSettingsLocked(SettingsProvider.SETTINGS_TYPE_SECURE, userId);
+    }
+
+    private static void insertSetting(SettingsState state, String key, String value) {
+        if (state == null || key == null || value == null) {
```
muhomorr: It's better to crash here than to fail silently when `null` is passed.
In general, it's better to add null checks only when `null` is an expected value.

quh4gko8:
superseded by https://github.com/GrapheneOS/platform_frameworks_base/pull/388

PR #325, opened 2023-02-10
add a setting for forcibly disabling SUPL

muhomorr:


IzzySoft:
Alternatively, you could also establish a [SUPL server](https://www.tajuma.com/supl/) (like you have a captive-portal one) and allow to set that/set that by default/prompt for at setup. It would then "proxy" SUPL requests. Maybe cutting out the IMSI then as well (or fake a random one by scrambling the device-specific part) if that software doesn't do it already, as else it would be a bit meaningless.

Not sure if that's an option – if not, simply ignore me :wink:

PS: I have not tried that software yet, it was just brought to my attention recently. So it's just an idea.

thestinger:
SUPL can already be disabled via APN configuration but we found that it doesn't fully disable it on 6th/7th generation Pixels and we also found that SUPL can be used without a carrier when not using airplane mode so that's why we're adding a separate toggle from the existing APN types control.

We plan on hosting our own SUPL proxy similar to our other hosted services. We provide our own connectivity check server, secure time server (HTTPS-based instead of NTP until Android adds NTS), PSDS cache and hardware attestation remote provisioning proxy:

https://grapheneos.org/faq#default-connections

SUPL should be easy enough to add to our existing setup. We also already disable sending IMSI and phone number to the SUPL server.

IzzySoft:
Thanks Daniel! Looking forward to your SUPL Server! :star_struck:

thestinger:
We also already have our own DNS-over-TLS check service which is simply provided via https://github.com/GrapheneOS/ns1.grapheneos.org/blob/main/zones.yaml#L321-L326. We treat that as related to the connectivity checks. Instead of a separate toggle, we resolve random subdomains of our DNS check address when using our connectivity check service and resolve random subdomains of the standard Google domain when using the standard connectivity checks. That's covered in https://grapheneos.org/faq#default-connections.

It shouldn't be hard to host our own SUPL service but it's a pain to test especially on Qualcomm devices where the baseband makes the requests to it rather than the OS. The OS configures it but we can't easily debug it from the OS there. Can easily see if it's working or not based on how fast you get a GNSS lock after reboot though.

We now have a SUPL TLS proxy, although it hasn't yet been tested with the OS:

https://github.com/GrapheneOS/grapheneos.network/commit/6e08b00e9e52087a0c37cd7b4bf05d2279b4cc6d
https://github.com/GrapheneOS/ns1.grapheneos.org/commit/c261f98988ab52435996b529a80fef2ba8b85bff
https://github.com/GrapheneOS/infrastructure/commit/8b96ee620ced24db944403a041e39f1393067134

We'll be removing our force disable toggle for SUPL and replacing it with a 3 state toggle between GrapheneOS, Standard (carrier/fallback) and Disabled where GrapheneOS is the default and overrides the host/port to point to our service, but won't force enable it if the carrier doesn't use SUPL.

thestinger:
We now have a SUPL TLS proxy, although it hasn't yet been tested with the OS

https://github.com/GrapheneOS/ns1.grapheneos.org/commit/c261f98988ab52435996b529a80fef2ba8b85bff
https://github.com/GrapheneOS/grapheneos.network/commit/79971347aaaec1127afda9061870ca1b0e0fda9b
https://github.com/GrapheneOS/infrastructure/commit/8b96ee620ced24db944403a041e39f1393067134

We'll be removing our force disable toggle for SUPL and replacing it with a 3 state toggle between GrapheneOS, Standard (carrier/fallback) and Disabled where GrapheneOS is the default and overrides the host/port to point to our service, but won't force enable it if the carrier doesn't use SUPL.

thestinger:
We now have a SUPL TLS proxy, although it hasn't yet been tested with the OS

https://github.com/GrapheneOS/ns1.grapheneos.org/commit/c261f98988ab52435996b529a80fef2ba8b85bff
https://github.com/GrapheneOS/grapheneos.network/commit/ec54bca065eabeef6a88b7576bec20b059afe80f
https://github.com/GrapheneOS/infrastructure/commit/8b96ee620ced24db944403a041e39f1393067134

We'll be removing our force disable toggle for SUPL and replacing it with a 3 state toggle between GrapheneOS, Standard (carrier/fallback) and Disabled where GrapheneOS is the default and overrides the host/port to point to our service, but won't force enable it if the carrier doesn't use SUPL.

thestinger:
We now have a SUPL TLS proxy, although it hasn't yet been tested with the OS:

https://github.com/GrapheneOS/grapheneos.network/commit/dff2401254193b0c420c66545a05fae3e00562f4
https://github.com/GrapheneOS/ns1.grapheneos.org/commit/c261f98988ab52435996b529a80fef2ba8b85bff
https://github.com/GrapheneOS/infrastructure/commit/8b96ee620ced24db944403a041e39f1393067134

We'll be removing our force disable toggle for SUPL and replacing it with a 3 state toggle between GrapheneOS, Standard (carrier/fallback) and Disabled where GrapheneOS is the default and overrides the host/port to point to our service, but won't force enable it if the carrier doesn't use SUPL.

thestinger:
Due to the TLS PSK system used by SUPL, using nginx as the reverse proxy didn't work. It's possible we can get it working using stunnel or a similar service but it's more of a mess than expected since it derives a PSK (pre-shared key) from the domain which is not something nginx supports. It's also possible it uses the PSK within the tunnel too.

thestinger:
We've successfully gotten the SUPL proxy working with both Qualcomm SUPL and Broadcom SUPL. Information available at https://grapheneos.social/@GrapheneOS/109960680764836371. Our proxy is used to override the carrier-chosen SUPL server by default in the upcoming GrapheneOS release. Users can choose to instead use the carrier SUPL server or disable the feature. We could take a similar approach to geocoding and network-based location but we want to provide the option to use local databases for those too. SUPL is similar to a network location service in a lot of ways but is designed to simply accelerate GNSS and it's only queried very infrequently for bootstrapping to get a very approximate location. It gives a lot less data to the service than a network location service and since it's infrequent only needs short lived connections which mixes well with our privacy preserving proxy since clients just make a single query and then don't usually reconnect for a long time. Qualcomm even caches the result between reboots so rebooting doesn't cause a new query. Broadcom will query again after reboot.

IzzySoft:
@thestinger that's great news, thanks Daniel! Will you be open to the use of your SUPL proxy by "other devices" (i.e. those not running GOS), as you are with your captive-portal check? You mention http://supl.grapheneos.org/ yourself in a toot, so I assume as much. That would be wonderful!

thestinger:
You can use it as an individual but software shouldn't be redistributed to others using our services without our permission since users choosing to use it themselves would be an insignificant cost for us but a software project using it for many users could significantly increase our load.

IzzySoft:
Fair enough, thanks! So software may offer it as an option, but it should not be enabled by default.

thestinger:
Yes, that would be fine. If other OSes want to enable it by default they should ask us. If they have significant revenue they should help fund our infrastructure. We publish most of our server configuration so it's easy enough to self-host this anyway.

thestinger:
Connectivity checks, HTTPS network time (similar to connectivity checks but with a millisecond precision X-Time header and the client warms up the connection first), PSDS cache, attestation key provisioning proxy and SUPL proxy are all provided by the grapheneos.network nginx configuration:

https://github.com/GrapheneOS/grapheneos.network/blob/main/nginx/nginx.conf

On current generation devices those are most of the normal default connections made by the OS other than the releases.grapheneos.org server for OS updates and app updates via our app repository system. DNS-over-TLS checks happen via the DNS resolver server so it's the ISP DNS server or user configured DNS server connecting to our nameservers (or Google's nameservers). We currently only use our PSDS cache for Broadcom GNSS devices (Tensor Pixels) and don't change the server for Qualcomm GNSS devices (Snapdragon Pixels). We could add Qualcomm PSDS cache too but it's not entirely clear how to implement it cleanly since the URLs are set in a configuration file read by the GPS service and so we might need to have 2 configuration files and restart the service with a different configuration which is a lot different than Broadcom where they use AOSP PSDS downloader.

IzzySoft:
Thanks for sharing those details, too!

mherrmann3:
Great news & big thanks for this effort!
And a +1 for the local database.

But I guess using the proxy on any other OS (like LineageOS) is snakeoil, since the SUPL request, which is forwarded to a Google server, naturally includes the IMSI/phone number.[^1] Or how [Marvin W](https://github.com/mar-v-in) put's it in [a Mastodon post](https://mastodon.social/@larma/109846110714861908) (translated from German): 
> _"[This proxy] mainly prevents that if someone intercepts the network traffic it says "Connection to Google", but then just "Connection to GrapheneOS Server"._

Is that correct? Thanks for clarifying.

[^1]: LineageOS [abandoned the patch to remove it](https://review.lineageos.org/c/LineageOS/android_frameworks_base/+/249219) unlike GrapheneOS, which [implemented that patch](https://github.com/GrapheneOS/platform_frameworks_base/pull/111/commits/26ff3944b0ab12fb1e51b0edf947d91240278269)

thestinger:
They might be missing that we already disabled sending the IMSI and phone number. The reason to mask the IP address is because the mobile IP address will remain the same as you move around and will end up providing multiple locations tied to a single IP address over time. Since SUPL connections are short lived, our proxy serves to separate each request so that the SUPL server only gets a location without an identifier. Our SUPL proxy **could** zero out IMSI and IP address but we disabled that client side for GrapheneOS which is superior. One benefit of the proxy is now we can test that it's actually working as intended especially for the Qualcomm SUPL case done by the baseband where we previously had no way to determine if disabling IMSI and phone number was working.

thestinger:
We also upgraded TLS security. We're planning on submitting that part upstream.

thestinger:
For key provisioning, Google says they use a reverse proxy like ours themselves where they handle the connections separately from decrypting the messages to handle key provisioning itself, without tying the data together. In that case, our proxy avoids needing to trust that Google is doing that correctly/securely since there are essentially 2 proxies and our proxy is unable to see the decrypted data. If either ours or theirs got compromised, not both, someone couldn't tie those things together. SUPL proxy is not quite that useful but it's still useful. It's inherently not possible for us to host key provisioning ourselves for devices that weren't made to be GrapheneOS devices with a GrapheneOS key attestation root. For SUPL, we can eventually host our own, but I'm not aware of any software for doing it. We'd just need to add an extra subdomain and an extra choice to use our service instead of our proxy. We know how to handle TLS, etc. for it now and we know it's working for both Broadcom and Qualcomm SUPL so a lot of progress has been made.

thestinger:
IP address can be tied to other things. Broadcom SUPL will go through a VPN but Qualcomm SUPL is done by the baseband and always goes over the mobile network. Carrier likely has a persistent record of which IPv4 (likely shared) and IPv6 address (likely unique) belonged to each user a given time. I don't think it's fair to claim that there is no benefit from the proxy. IP address is not sensitive by itself when all it reveals is that the device uses an Android-based OS but IP address + location data becomes problematic. Some other operating systems use supl.vodafone.com and if you check the DNS resolution for that it's just a CNAME for supl.google.com. I'm not even aware of another SUPL service that we could offer as an option to users. We **could** provide the ability to enter an arbitrary domain as a string or list more alternatives but what good is that if there aren't any working options available? supl.vodafone.com doesn't actually work if you do TLS verification like both Qualcomm and Broadcom SUPL do by default since supl.google.com doesn't provide that:

> curl: (60) SSL: no alternative certificate subject name matches target host name 'supl.vodafone.com'

It is not like connectivity checks and network time which are just connecting and not submitting anything. We replaced unauthenticated NTP with HTTPS time for security originally rather than to avoid a Google or far less untrustworthy third party (NTP pool) service anyone can join. PSDS is also just fetching data and despite us implementing that with nginx proxy functionality, we always serve them from the cache and it updates the cache in the background. It will serve previous stale data if Broadcom's service doesn't work.

IzzySoft:
> naturally includes the IMSI/phone number

AFAIK that's also sanitized by e.g. ShiftOS (SHIFTPHONES is a German OEM) and others who applied a patch to "remove sensitive info from SUPL requests". The other part (network request) was already answered in detail by Daniel (Google won't see the device IP anymore but just the proxy IP, so more privacy).

> I'm not even aware of another SUPL service that we could offer as an option to users.

I've never heard about any other either (just about the very same Vodafone "fake", which for obvious reasons does not count).

thestinger:
I've seen configurations indicating that various other SUPL services used to exist from various phone OEMs (Nokia) and carriers but none of the ones that I've seen exist anymore. It also seems that most carriers didn't bother with control plane A-GNSS and instead rely on SUPL via supl.google.com. Broadcom GNSS doesn't seem to support control plane A-GNSS, only SUPL, since it's not closely tied to the baseband like Qualcomm SUPL which is capable of doing control plane A-GNSS instead.

The idea seems to have been that carriers would mostly do control plane A-GNSS and carriers would host SUPL if they couldn't deploy that easily. I don't think it was expected that it would be delegated to another service provider with knowledge of cell towers across carriers instead of the carrier providing the service. It ended up this way from carriers being very lazy.

thestinger:
Also, AOSP currently seems to artificially limit SUPL usage to when cellular is available but Broadcom SUPL could work without a cellular connection when the device isn't in airplane mode since it only needs to be able to scan for nearby cell towers. Qualcomm SUPL can only connect via mobile data since it's specifically part of the cellular baseband. Qualcomm's baseband also handles Wi-Fi and Bluetooth but it's a soft MAC Wi-Fi implementation mostly done by the OS and it can't use it. Broadcom SUPL is regular OS internet traffic. It will use any available connection (Wi-Fi, Ethernet, etc.) and will go through the Owner user's VPN.

mherrmann3:
Thanks, Daniel for your detailed write up and sharing your experience & knowledge.

I understood that your proxy is (very) useful for GrapheneOS users. But for those with other OS (except ShiftOS, apparently—thanks, Izzy), the included IMSI/phone number makes the user still unique—even more unique than the (now-masked) IP address—so that's why I had asked for the proxy's usefulness for non-GrapheneOS users.
But I didn't understand the key provisioning / attestation / TLS part in this context—so the proxy wouldn't even work with other OS?

I also had come across a few SUPL servers; most are now offline, but two of them are still ping-able. But I'm not sure if they respond with a useful SUPL answer.
* `agps.fetnet.net`
* `lbs.geo.t-mobile.com` (perhaps requires TLS rootcert [[github.com/Astralix/Android_gps/.../gps.conf](https://github.com/Astralix/Android_gps/blob/master/gps.conf)])

Since you mention control plane ... I guess there is no way to check if/when the Qualcomm baseband uses it. Or is there some characteristic string in the logcat?

Btw: SUPL servers could in principle also work in MSA (Mobile Station Assisted) mode (calculating the position directly using various information), but Google's SUPL doesn't support that [[Wikipedia](https://en.wikipedia.org/wiki/Assisted_GNSS)], so I wonder if a full-fledged SUPL server ever existed? I guess you already answered it: _carriers being very lazy_. 😢

thestinger:
> But I didn't understand the key provisioning / attestation / TLS part in this context—so the proxy wouldn't even work with other OS?

I was explaining the other services implemented by our network servers. Attestation key provisioning server inherently has to be a proxy to Google's key attestation provisioning server for devices shipped using the Google key attestation root in their hardware keystores. If there were devices using a Huawei key attestation root, we'd need a separate proxy for those. We'd only be able to have our own key attestation server for devices built to run GrapheneOS using a GrapheneOS key attestation root.

> I had come across a few SUPL servers; most are now offline, but two of them are still ping-able. But I'm not sure if they respond with a useful SUPL answer.

Neither responds to a TLS connection on port 7275. Compare `openssl s_client -connect supl.google.com:7275` with `openssl s_client -connect lbs.geo.t-mobile.com:7275`. Maybe they only accept connections via cellular data within the relevant carriers.

> Since you mention control plane ... I guess there is no way to check if/when the Qualcomm baseband uses it. Or is there some characteristic string in the logcat?

It's possible to capture the traffic to analyze, just difficult, since without an exploit you need your own base station / SIM setup.

It's quite possible detailed logs can be obtained via the baseband debugging / diagnostic features but they aren't enable in typical production builds.

> Btw: SUPL servers could in principle also work in MSA (Mobile Station Assisted) mode (calculating the position directly using various information)

Android seems to support it, so perhaps there were implementations.

thestinger:
> I understood that your proxy is (very) useful for GrapheneOS users. But for those with other OS (except ShiftOS, apparently—thanks, Izzy), the included IMSI/phone number makes the user still unique—even more unique than the (now-masked) IP address—so that's why I had asked for the proxy's usefulness for non-GrapheneOS users.

Still seems useful. User installed apps, websites, etc. can obtain IP address but not IMSI. Apps can only get phone number if Phone permission is granted. If in theory Google's server was compromised and had data leaked or they gave the information to a government based on a demand it could be cross-referenced with IP data, etc.

mherrmann3:
FYI: LineageOS now also removes IMSI / phone number from SUPL requests, **unless** in emergency: [review.lineageos.org/c/LineageOS/android_frameworks_base/+/350287](https://review.lineageos.org/c/LineageOS/android_frameworks_base/+/350287/).

The patch is based on [review.calyxos.org/c/CalyxOS/platform_frameworks_base/+/15533](https://review.calyxos.org/c/CalyxOS/platform_frameworks_base/+/15533) (see also recent update in [CalyxOS Issue#618](https://gitlab.com/CalyxOS/calyxos/-/issues/618)).

thestinger:
@mherrmann3 That change only disables providing IMSI and still provides the phone number.

thestinger:
@mherrmann3 I don't understand why they're doing it that way instead of testing it as we did and then making a decision based on confirmation that the change works as intended and doesn't hinder the feature. We still need to test the Snapdragon case to see if this is even relevant to it. This doesn't necessarily have an impact for all SUPL implementations, since the Snapdragon one is capable of retrieving this information itself due to being done in a baseband RTOS process. Our proxy enables us to easily test this which was previously not easy because the baseband makes the SUPL request via TLS using mobile data and does certificate verification so even if you can intercept mobile data, TLS prevents you seeing the data going back and forth without configuring it to connect to your own server.

thestinger:
As I mentioned above in https://github.com/GrapheneOS/platform_frameworks_base/pull/325#issuecomment-1455121300, our proxy can be modified to zero out the IMSI and phone number data. We could use the nginx njs module for this, although I haven't used it before and would need to figure out how to use it. We also currently don't know the format of the SUPL messages and would need to figure that out from the standard. For testing, sending data with IDs enabled and then sending data without IDs enabled followed by comparing the results can work but not for zeroing it which always needs to be done precisely.

PR #326, opened 2023-02-11
add support for user profile switch on lockscreen

ghost:
This feature was added during Android 13 DP1 but never shown up in the final build.

The underlying support is there and allows users to switch User Profiles from the keyguard. You can do this anyway from the lockscreen (swipe down > click on user profile picture > select your profile).

[Link to original XDA Article](https://www.xda-developers.com/android-13-switch-user-profiles/)

Picture from emulator
image


x86pup:
> This feature was added during Android 13 DP1 but never shown up in the final build.

It's very likely it's because it's not finished. There are various hidden features in AOSP that are disabled by default because they aren't finished like the lock screen clock theme picker.

ghost:
> > This feature was added during Android 13 DP1 but never shown up in the final build.
> 
> It's very likely it's because it's not finished. There are various hidden features in AOSP that are disabled by default because they aren't finished like the lock screen clock theme picker.

Would agree with that. Is there any tests I can run through? It seems to work with switching profiles and guests.

PR #331, opened 2023-02-18
migrate PIN scrambling to ExtSettings

x86pup:
Depends on https://github.com/GrapheneOS/platform_packages_apps_Settings/pull/146


muhomorr:
I think it's better not to revert the original commit in this case, most of the code remains the same.

x86pup:
Merged in https://github.com/GrapheneOS/platform_frameworks_base/pull/345

PR #333, opened 2023-02-19
add unlocked bootloader warning

x86pup:
Currently untested on an actual Pixel, emulator does not support these properties.


x86pup:
Possibly solves https://github.com/GrapheneOS/os-issue-tracker/issues/576 depending on whether it's wanted in SetupWizard or not.

PR #335, opened 2023-02-24
make screenshot sound use touch sounds setting

x86pup:
Removes screenshot sound instead of adding a toggle

### DIFF
--- a/packages/SystemUI/src/com/android/systemui/screenshot/ScreenshotController.java
+++ b/packages/SystemUI/src/com/android/systemui/screenshot/ScreenshotController.java
@@ -889,6 +889,10 @@ public class ScreenshotController {
     }
 
     private void playCameraSound() {
+        if (Settings.System.getInt(mContext.getContentResolver(), SOUND_EFFECTS_ENABLED, 0) != 1) {
+            return;
+        }
+
         mCameraSound.addListener(() -> {
             try {
                 MediaPlayer player = mCameraSound.get();


x86pup:
Solves https://github.com/GrapheneOS/os-issue-tracker/issues/1696

muhomorr:
Can the screenshot capture sound be removed entirely instead?

thestinger:
That's an option but I'm not sure what is best.

muhomorr:
There's a prominent visual indication after screenshot is captured.
If TalkBack is enabled, there's a "Saving screenshot" announcement.

x86pup:
I can look into removing it. Do you want just the screenshot sound method call removed like this? Or do you want the entirety of it removed? The latter will make it harder to port in the future though and not sure if it will break anything else.

muhomorr:
There's one more caller of `playCameraSound()`.
Simpler to just remove contents of `playCameraSound()`.

x86pup:
Do we also want to remove the visual indication of a screenshot?

muhomorr:
I think not. I meant that the sound is redundant, since there's both a visual indication and a TalkBack announcement.

x86pup:
Removed screenshot sound instead, tested on my Pixel 5 screenshot sound is indeed gone.

muhomorr:
I think better to remove `playCameraSound()` method too, it'll make sure build will fail in case there's more callers of it in the future.
Alternatively, can just remove contents of `playCameraSound()`.

x86pup:
@muhomorr Removed the contents of playCameraSound() instead

muhomorr (approved):

packages/SystemUI/src/com/android/systemui/screenshot/ScreenshotController.java
```diff
@@ -854,8 +854,11 @@ private void playCameraSound() {
      * failure).
      */
     private void saveScreenshotAndToast(Consumer<Uri> finisher) {
-        // Play the shutter sound to notify that we've taken a screenshot
-        playCameraSound();
+        if (Settings.System.getInt(mContext.getContentResolver(),
```
muhomorr: Better to move the check inside `playCameraSound()`

x86pup: Moved the check

packages/SystemUI/src/com/android/systemui/screenshot/ScreenshotController.java
```diff
@@ -838,15 +838,18 @@ private ListenableFuture<MediaPlayer> loadCameraSound() {
     }
 
     private void playCameraSound() {
-        mCameraSound.addListener(() -> {
-            try {
-                MediaPlayer player = mCameraSound.get();
-                if (player != null) {
-                    player.start();
+        if (Settings.System.getInt(mContext.getContentResolver(),
```
muhomorr: Better to return early instead, it'll likely simplify porting.

x86pup: Made it return early if the setting is not true/enabled

PR #353, opened 2023-03-28
Configurable clipboard autoclear timeout support

secrecyflag:
Dependent on: https://github.com/GrapheneOS/platform_packages_apps_Settings/pull/158
Needs testing!


core/java/android/provider/Settings.java
```diff
@@ -17330,6 +17330,12 @@ public static boolean putFloat(ContentResolver cr, String name, float value) {
          */
         public static final String BLUETOOTH_OFF_TIMEOUT = "bluetooth_off_timeout";
 
+        /**
+         * The amount of time in milliseconds before the clipboard is cleared.
+         * @hide
+         */
+        public static final String CLIPBOARD_AUTO_CLEAR_TIMEOUT = "clipboard_auto_clear_timeout";
```
x86pup: ExtSettings infrastructure does not touch Settings.java, you need to define this in [ExtSettings.java](https://github.com/GrapheneOS/platform_frameworks_base/blob/13/core/java/android/ext/settings/ExtSettings.java) instead

services/core/java/com/android/server/clipboard/ClipboardService.java
```diff
@@ -429,9 +429,7 @@ private void scheduleAutoClear(@UserIdInt int userId, int intendingUid) {
         }
 
         private long getTimeoutForAutoClear() {
-            return DeviceConfig.getLong(DeviceConfig.NAMESPACE_CLIPBOARD,
-                    PROPERTY_AUTO_CLEAR_TIMEOUT,
-                    DEFAULT_CLIPBOARD_TIMEOUT_MILLIS);
+            return Settings.Global.getLong(mContext.getContentResolver(), Settings.Global.CLIPBOARD_AUTO_CLEAR_TIMEOUT, DEFAULT_CLIPBOARD_TIMEOUT_MILLIS);
```
x86pup: ExtSettings checks should import `import android.ext.settings.ExtSettings;` and do:

`ExtSettings.<key>.get(<context>)`

In this case:

`return ExtSettings.CLIPBOARD_AUTO_CLEAR_TIMEOUT.get(mContext)`

As we're not using Settings.java.

core/java/android/ext/settings/ExtSettings.java
```diff
@@ -9,6 +9,7 @@
 
 /** @hide */
 public class ExtSettings {
+    public static final int DEFAULT_CLIPBOARD_TIMEOUT_MILLIS = 3600000;
```
x86pup: ExtSettings constants should be defined in their own file, but in this case I don't think this needs to be a constant. In fact, we could just use the public constant from frameworks/base/services/core/java/com/android/server/clipboard/ClipboardService.java

`import static com.android.server.clipboard.ClipboardService.DEFAULT_CLIPBOARD_TIMEOUT_MILLIS`

core/java/android/ext/settings/ExtSettings.java
```diff
@@ -31,6 +32,9 @@ public class ExtSettings {
     public static final BoolSetting SCRAMBLE_PIN_LAYOUT = new BoolSetting(
             Setting.Scope.PER_USER, "lockscreen_scramble_pin_layout", false);
 
+    public static final IntSetting CLIPBOARD_AUTO_CLEAR_TIMEOUT = new IntSetting(
+            Setting.Scope.GLOBAL, "clipboard_auto_clear_timeout", DEFAULT_CLIPBOARD_TIMEOUT_MILLIS);
```
x86pup: I think we also want this to be per-user instead of global

x86pup (changes requested):
Needs some changes, needs to be rebased, and I think we want this to be per-user instead of global unless due to some limitation.

I did type these suggestions on GitHub which may be incorrect / hard to understand, so if you want I can push my requested changes to my fork.

core/java/android/ext/settings/ExtSettings.java
secrecyflag: I tried that, but for some reason I couldn't figure out, `make` complains that `com.android.server.clipboard` doesn't exist?? I'm not sure if the package should be added to `Android.bp`, but I left it for now this way. Could you verify it?

x86pup: Shouldn't need to touch Android.bp, but I can look at it later today

secrecyflag: Solved by moving the constant to a new file.

secrecyflag:
Should be ready for testing on a pixel.

x86pup:
Still needs to be rebased

x86pup:
Toggle doesn't seem to show up in my Pixel 6 after building.

(See review in Settings for solution)

services/core/java/com/android/server/clipboard/ClipboardService.java
```diff
@@ -429,9 +430,7 @@ private void scheduleAutoClear(@UserIdInt int userId, int intendingUid) {
         }
 
         private long getTimeoutForAutoClear() {
```
x86pup: Seems like this isn't a per-user Context (and mContext seems invalid here, build error) which will cause the settings in the Owner profile to apply across all profiles irrespective of the current profile's setting, so you'll need to specify the uid this is for.

```suggestion
        private long getTimeoutForAutoClear(int userId) {
            return ExtSettings.CLIPBOARD_AUTO_CLEAR_TIMEOUT.get(getContext(), userId);
```

And where `getTimeoutForAutoClear` method is called, pass the `userId` argument:

`getTimeoutForAutoClear(userId)`

x86pup (changes requested):
Need to replace `mContext` with `getContext()` method, and get the userId as the Context is not per-user

services/core/java/com/android/server/clipboard/ClipboardService.java
secrecyflag: Didn't catch that build error as I'm currently using `m Settings`. Waiting for a new SSD to arrive (don't have enough space on device).

secrecyflag:
Closing this pull request as it became messy. Will open a new one and send here.

secrecyflag:
#357 @r3g-5z

PR #357, opened 2023-04-02
Configurable clipboard auto-clear timeout

secrecyflag:
solves: https://github.com/GrapheneOS/os-issue-tracker/issues/2012


core/java/android/ext/settings/ExtSettings.java
```diff
@@ -6,10 +6,11 @@
 import static android.ext.settings.GnssConstants.SUPL_DISABLED;
 import static android.ext.settings.GnssConstants.SUPL_SERVER_GRAPHENEOS_PROXY;
 import static android.ext.settings.GnssConstants.SUPL_SERVER_STANDARD;
+import static android.ext.settings.ClipboardConstants.DEFAULT_CLIPBOARD_TIMEOUT_MILLIS;
+
 
 /** @hide */
 public class ExtSettings {
-
```
x86pup: Add back this new line

core/java/android/ext/settings/ClipboardConstants.java
```diff
@@ -0,0 +1,9 @@
+package android.ext.settings;
+
+/**
```
x86pup: Just to stay consistent with other constants, could this comment be one line?:

```suggestion
/** @hide */
```

services/core/java/com/android/server/clipboard/ClipboardService.java
```diff
@@ -62,6 +62,7 @@
 import android.os.UserManager;
 import android.provider.DeviceConfig;
 import android.provider.Settings;
+import android.ext.settings.ExtSettings;
```
x86pup: Imports in majority of AOSP code are ordered alphabetically ascending (A to Z) from left to right, could this be moved to be right above `import android.graphics.drawable.Drawable;`?

```
import android.content.pm.UserInfo;
import android.ext.settings.ExtSettings;
import android.graphics.drawable.Drawable;
```

core/java/android/ext/settings/ExtSettings.java
```diff
@@ -6,10 +6,11 @@
 import static android.ext.settings.GnssConstants.SUPL_DISABLED;
 import static android.ext.settings.GnssConstants.SUPL_SERVER_GRAPHENEOS_PROXY;
 import static android.ext.settings.GnssConstants.SUPL_SERVER_STANDARD;
+import static android.ext.settings.ClipboardConstants.DEFAULT_CLIPBOARD_TIMEOUT_MILLIS;
+
```
x86pup: Remove this extra new line

x86pup (changes requested):
Just a few style changes to stay consistent and should be good from my POV.

core/java/android/ext/settings/ExtSettings.java
```diff
@@ -6,6 +6,7 @@
 import static android.ext.settings.GnssConstants.SUPL_DISABLED;
 import static android.ext.settings.GnssConstants.SUPL_SERVER_GRAPHENEOS_PROXY;
 import static android.ext.settings.GnssConstants.SUPL_SERVER_STANDARD;
+import static android.ext.settings.ClipboardConstants.DEFAULT_CLIPBOARD_TIMEOUT_MILLIS;
```
x86pup: Move this above the GnssConstants for alphabetical ascending imports

x86pup (changes requested):
Slight change here

x86pup (approved):
Concerns have been addressed and feature works

Must be squashed when merged

secrecyflag:
will rebase today.

secrecyflag:
re-based successfully. the feature should be ready now.

core/java/android/ext/settings/ClipboardConstants.java
```diff
@@ -0,0 +1,7 @@
+package android.ext.settings;
+
+/** @hide */
+public class ClipboardConstants {
+    // From com.android.server.clipboard.ClipboardService.DEFAULT_CLIPBOARD_TIMEOUT_MILLIS 
+    public static final int DEFAULT_CLIPBOARD_TIMEOUT_MILLIS = 3600000;
```
muhomorr, line 6: There's no need to declare this as a separate constant in a separate file.
It's better to use `TimeUnit.HOURS.toMillis(1)` in `CLIPBOARD_AUTO_CLEAR_TIMEOUT` definition (default value can likely be reduced to several minutes) and to remove constant `com.android.server.clipboard.ClipboardService.DEFAULT_CLIPBOARD_TIMEOUT_MILLIS` to make sure it's not used elsewhere.

secrecyflag: This is just following the convention from `com.android.server.clipboard.ClipboardService.DEFAULT_CLIPBOARD_TIMEOUT_MILLIS` for making it easier to port to other versions of AOSP.

secrecyflag: This also includes changing CTS as this value is used there. Depends if you don't mind.

muhomorr: If constant is removed, build will fail if it's used elsewhere in a new AOSP version. This is more reliable than assuming that search for usages of this constant will be performed each time new AOSP version comes out.

muhomorr: Another option is to write to DeviceConfig setting directly instead of replacing it with a new setting.

secrecyflag: Why would you want to use DeviceConfig? Isn't the point of ExtSettings is to easily add new settings? This constant (`ClipboardConstants.DEFAULT_CLIPBOARD_TIMEOUT_MILLIS`) is not used elsewhere, only in `ExtSettings`, thus making it easier to port.

muhomorr: I'm referring to `com.android.server.clipboard.ClipboardService.DEFAULT_CLIPBOARD_TIMEOUT_MILLIS` constant, not to `ClipboardConstants.DEFAULT_CLIPBOARD_TIMEOUT_MILLIS`

muhomorr: Why add a new setting when AOSP already has a setting for the same thing?

secrecyflag: Oh yes but we tried to use it and it can't be imported (`ClipboardService.DEFAULT_CLIPBOARD_TIMEOUT_MILLIS`). You mean adding it to `DeviceConfig` as a setting?

muhomorr: I mean that the setting is already present in AOSP, it just doesn't have a UI:
https://github.com/GrapheneOS/platform_frameworks_base/blob/71a9bb51861736106919e97afec14a9bccdc4e10/services/core/java/com/android/server/clipboard/ClipboardService.java#L432

secrecyflag: I understand. Then instead of using `ExtSettings.CLIPBOARD_AUTO_CLEAR_TIMEOUT` in the [controller](https://github.com/GrapheneOS/platform_packages_apps_Settings/pull/158/commits/70dd2301f20507fe30154e035b958d890747fac2) we would use `DeviceConfig`. But we would need now to implement `AbstractListPreferenceController` because `IntSettingPrefController` uses `IntSetting`.

secrecyflag: Using `DeviceConfig` can be a bit problematic since we need to use some members from `ClipboardService` in [`ClipboardTimeoutController`](https://github.com/GrapheneOS/platform_packages_apps_Settings/pull/158/commits/70dd2301f20507fe30154e035b958d890747fac2) (and `com.android.server.clipboard` can't be imported.) A fixable solution is to use `ClipboardManager` instead but we would need to add a few methods that use the (needed) members from `ClipboardService`.

secrecyflag: `DeviceConfig` requires `PROPERTY_AUTO_CLEAR_TIMEOUT`, and `DEFAULT_CLIPBOARD_TIMEOUT_MILLIS` (if we want consistency) from `ClipboardService`. It's possible with a few changes.

muhomorr: Can just move these fields to ClipboardManager.

secrecyflag: How would we make it per-user? The current implementation doesn't allow per-user configuration because it uses `DeviceConfig`.

muhomorr: I don't think this setting needs to be per-user.

secrecyflag: It adds better isolation, and even security (no leaks between user profiles).

muhomorr: DeviceConfig settings are unreadable by unprivileged apps.

I've thought some more about this feature, and I think that it's better to just reduce this timeout to 10 or 5 minutes instead of adding one more setting. There's a large number of privacy-sensitive timeouts and flags in AOSP, making them configurable is not a right approach in most cases in my opinion.

secrecyflag: I believe it's still a good feature to add. People can copy passwords (for some reason) or other sensitive information from other apps and not worry about it existing for a long time and some app stealing it. It's also a pretty small change. Just a "setting" file and copying two constants. Can be easily added.

muhomorr:
I stand by what I wrote previously in the review: 

> I think that it's better to just reduce this timeout to 10 or 5 minutes instead of adding one more setting. There's a large number of privacy-sensitive timeouts and flags in AOSP, making them configurable is not a right approach in most cases in my opinion.

secrecyflag:
If so, we should use a small timeout like one minute. An enhancement would be resetting the timeout every time the user pastes the same clipboard.

muhomorr:
I think one minute is way too short.

Another approach would be to deny read access to clipboard for all apps (except IME) by default, and show a notification when an app tries to access it.

Basically, an off-by-default per-app toggle for clipboard access.

secrecyflag:
Yeah one minute is short that's why I thought every time the user pastes the clipboard it would reset. But a clipboard access permission is also possible and also be easily implemented. I think both (one minute timeout or more, and...) of the approaches are good enhancements.

muhomorr:
I think that one minute is too short even for the first paste.

secrecyflag:
Right it doesn't have to be one minute, I'm only clarifying that both of the approaches are good. It could be even five minutes.

PR #358, opened 2023-04-04
prevent Play services adding Wallet tile by itself

x86pup:
the Wallet tile is a system tile that checks if NFC is supported on the device and if there's a registered default NFC payment provider using the secure setting "nfc_payment_default_component". if both of these are true, then the wallet tile will be reported as available to be added.

the tile can still be added manually and is still fully functional.

some users get confused or weirded out that this happens, and Google Pay / Wallet for NFC payments do not work on GrapheneOS at the moment so lets just make them happy.

before installing sandboxed Google Play
```
oriole:/ # settings list secure | grep -i NFC_PAYMENT_DEFAULT_COMPONENT
nfc_payment_default_component=null
```

after:
```
oriole:/ # settings list secure | grep -i NFC_PAYMENT_DEFAULT_COMPONENT
nfc_payment_default_component=com.google.android.gms/com.google.android.gms.tapandpay.hce.service.TpHceService
```


thestinger:
I don't want to special case this here.

Also, Java string comparisons need to use the `equals` method since `==` does reference quality for strings. For arrays, `equals` also does reference equality (an early oversight they couldn't ever fix) so there's `java.util.Arrays.equals`.

PR #376, opened 2023-06-12
Use True Black background for Settings again

woshiahmedc:
Use true black for dark mode for Settings. It already committed before as "commit c730588"

### DIFF
--- a/core/res/res/values/colors_device_defaults.xml
+++ b/core/res/res/values/colors_device_defaults.xml
@@ -23,7 +23,7 @@
     <color name="primary_device_default_settings_light">@color/primary_device_default_light</color>
     <color name="primary_dark_device_default_dark">@color/primary_device_default_dark</color>
     <color name="primary_dark_device_default_light">@color/primary_device_default_light</color>
-    <color name="primary_dark_device_default_settings">@color/primary_device_default_dark</color>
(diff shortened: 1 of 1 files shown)


thestinger:
Why? I think we should stick with following Material You.

woshiahmedc:
I thought true black background would be better for battery and eyes in oled screen

thestinger:
Black only makes a tiny difference compared to dark grey for battery life and it doesn't help here since this isn't an app users have open on the screen for hours. Material You is the overall theme for the OS and increasingly for apps, and I don't think it makes much sense to deviate from it when we want to move our own apps towards it and replace legacy AOSP apps partly to provide a consistent Material You theme.

woshiahmedc:
You're right, users don't open it for hours, I thought to apply true black for system-wide, this was a little part of that

thestinger:
We don't choose the theme for apps where users spend most of their time. The websites and apps choose their theme. Ideally, apps would all use Material You for a consistent theme everywhere. Some may decide to use a pure black background but others won't and it really makes very little difference from very dark grey for either battery life or eye strain.

woshiahmedc:
I understand, when I searched about true black mode in GrapheneOS, I saw it approved before while GrapheneOS using default dark theme now and I didn't know the reason, thank you for explanation

PR #388, opened 2023-07-03
Disable instant app features and adding users when locked upon SettingsState initialization

quh4gko8:


quh4gko8:
Merge with https://github.com/GrapheneOS/platform_packages_apps_Settings/pull/140

PR #396, opened 2023-07-26
[temporary] remove ScreenshotProxyService's dependency on CentralSurf…

octocorvus:
…aces

Workarounds issue that causes unexpected initialisation of OverviewProxyService, when a screenshot is taken, resulting in broken recents list and launcher crashes.

Test:
1. Switch to secondary user profile (which is at rest).
2. Take screenshot and press any of the buttons in the popup.
3. Open recents list and make sure that it is working properly and the launcher isn't crashing.


octocorvus:
superseded by https://github.com/GrapheneOS/platform_frameworks_base/pull/398

PR #445, opened 2023-11-11
add tests for connectivity checks setting

maade93791:


tests/GrapheneOS/ConnectivityCheckSettingTest/src/grapheneos/test/ConnectivityCheckSettingTest.java
```diff
@@ -0,0 +1,143 @@
...
+                        break;
+                    }
+                }
+            }
+            return result;
+        }
+    }
+
+    private final boolean checkParsed(ArrayList<String> matched, int setting) {
+        if (matched.size() <= 2) {
```
muhomorr: This check is incorrect for `GRAPHENEOS_SERVER_SETTING` and `STANDARD_SERVER_SETTING` (list size should be exactly 2).

Also, checks like these should use an assertion instead of a conditional to make test failures easier to debug.

e.g.
`assertEquals(matched.size(), 2)`

tests/GrapheneOS/ConnectivityCheckSettingTest/src/grapheneos/test/ConnectivityCheckSettingTest.java
```diff
@@ -0,0 +1,143 @@
...
+
+    private boolean baseTest(int setting) throws Exception {
+        toggleWifi("disabled");
+        getDevice().setProperty(SYS_PROP, Integer.toString(setting));
+        RunUtil.getDefault().sleep(1000); // 1s
+        getDevice().clearLogcat();
+        toggleWifi("enabled");
+        RunUtil.getDefault().sleep(10000); // 10s
+        
+        return getDevice().checkConnectivity();
```
muhomorr: Should use `assertTrue(getDevice().checkConnectivity())`

tests/GrapheneOS/ConnectivityCheckSettingTest/src/grapheneos/test/ConnectivityCheckSettingTest.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    @Test
+    public void testGrapheneOSServer() throws Exception {
+        assertTrue(baseTest(GRAPHENEOS_SERVER_SETTING));
+        assertTrue(checkParsed(parseLogcat(FILTER_REQ_REGEXP, GRAPHENEOS_SERVER_SETTING), GRAPHENEOS_SERVER_SETTING));
```
muhomorr: I think `checkParsed()` should be inlined here and in other tests to simplify the logic.

muhomorr (changes requested):

tests/GrapheneOS/ConnectivityCheckSettingTest/src/grapheneos/test/ConnectivityCheckSettingTest.java
```diff
@@ -0,0 +1,147 @@
...
+        getDevice().executeShellV2Command(cmd);
+    }
+
+    @Before
+    public void setUp() throws DeviceNotAvailableException {
+        if (!getDevice().isAdbRoot()) {
+            try {
+                getDevice().enableAdbRoot();
+            } finally {
+                fail("adb root access is required");
```
muhomorr, line 64: `finally` block runs unconditionally, which breaks the test if adb is not running as root

tests/GrapheneOS/ConnectivityCheckSettingTest/src/grapheneos/test/ConnectivityCheckSettingTest.java
```diff
@@ -0,0 +1,147 @@
...
+                        result.add(matcher.group(0));
+                        break;
+                    }
+                }
+            }
+            return result;
+        }
+    }
+
+    private final boolean checkParsed(ArrayList<String> matched, int setting) {
```
muhomorr: Why have this as a separate method? Each of switch cases is executed only once

tests/GrapheneOS/ConnectivityCheckSettingTest/src/grapheneos/test/ConnectivityCheckSettingTest.java
```diff
@@ -0,0 +1,147 @@
...
+            case DISABLED_SETTING -> matched.get(0).equals(FILTER_DISABLED_REGEXP);
+            default -> false;
+        };
+        return result;
+    }
+
+    private void baseTest(int setting) throws Exception {
+        toggleWifi("disabled");
+        getDevice().setProperty(SYS_PROP, Integer.toString(setting));
+        RunUtil.getDefault().sleep(1000); // 1s
```
muhomorr: Why sleep() here?

tests/GrapheneOS/ConnectivityCheckSettingTest/src/grapheneos/test/ConnectivityCheckSettingTest.java
```diff
@@ -0,0 +1,147 @@
...
+        getDevice().clearLogcat();
+        toggleWifi("enabled");
+        RunUtil.getDefault().sleep(10000); // 10s
```
muhomorr: This statement slows down the test massively, it's repeated for each of three cases. In most cases connectivity check completes much faster.

Better to use a loop:
```
boolean hasConnectivity = false;
for (int i = 0; i < 10; ++i) {
   sleep(1000)
   hasConnectivity = getDevice().checkConnectivity()
   if (hasConnectivity) {
      break
   }
}
assertTrue(hasConnectivity)
```

tests/GrapheneOS/ConnectivityCheckSettingTest/src/grapheneos/test/ConnectivityCheckSettingTest.java
maade93791: Sometimes `clearLogcat()` doesnt work properly. That results into incorrect logcat for the first test.

tests/GrapheneOS/ConnectivityCheckSettingTest/src/grapheneos/test/ConnectivityCheckSettingTest.java
```diff
@@ -0,0 +1,140 @@
...
+                Matcher matcher = pattern.matcher(line);
+                if (setting != DISABLED_SETTING) {
+                    if (matcher.find() && line.contains(LOG_TAG)) {
+                        result.add(line);
+                        if (result.toString().contains("PROBE_HTTP") && result.toString().contains("PROBE_HTTPS")) {
+                            break;
+                        }
+                    }
+                } else {
+                    if (matcher.find()) {
```
muhomorr: Missing check for `LOG_TAG`. `LOG_TAG` can be made part of regex.

tests/GrapheneOS/ConnectivityCheckSettingTest/src/grapheneos/test/ConnectivityCheckSettingTest.java
```diff
@@ -0,0 +1,140 @@
...
+        RunUtil.getDefault().sleep(1000);
+        assertTrue("device has no connectivity", hasConnectivity);
+    }
+
+    @Test
+    public void testGrapheneOSServer() throws Exception {
+        baseTest(GRAPHENEOS_SERVER_SETTING);
+        ArrayList<String> parsedLogcat = parseLogcat(FILTER_REQ_REGEXP, GRAPHENEOS_SERVER_SETTING);
+        assertEquals(parsedLogcat.size(), 2);
+        assertTrue(parsedLogcat.get(0).contains(ConnectivityCheckServer.GRAPHENEOS.httpUrl) && parsedLogcat.get(1).contains(ConnectivityCheckServer.GRAPHENEOS.httpsUrl));
```
muhomorr: Should use separate asserts for each condition, this simplifies finding test failure cause.

tests/GrapheneOS/ConnectivityCheckSettingTest/src/grapheneos/test/ConnectivityCheckSettingTest.java
```diff
@@ -0,0 +1,142 @@
...
+                }
+            }
+            return result;
+        }
+    }
+
+    private final long triggerCheck(int setting) throws Exception {
+        toggleWifi("disabled");
+        getDevice().setProperty(SYS_PROP, Integer.toString(setting));
+        getDevice().clearLogcat();
```
muhomorr: Is `clearLogcat()` necessary?

tests/GrapheneOS/ConnectivityCheckSettingTest/src/grapheneos/test/ConnectivityCheckSettingTest.java
```diff
@@ -0,0 +1,142 @@
...
+        assertEquals(parsedLogcat.size(), 2);
+        assertTrue(parsedLogcat.get(0).contains(ConnectivityCheckServer.GRAPHENEOS.httpUrl));
+        assertTrue(parsedLogcat.get(1).contains(ConnectivityCheckServer.GRAPHENEOS.httpsUrl));
+    }
+
+    @Test
+    public void testStandardServer() throws Exception {
+        long logStart = triggerCheck(STANDARD_SERVER_SETTING);
+        ArrayList<String> parsedLogcat = parseLogcat(FILTER_REQ_REGEXP, STANDARD_SERVER_SETTING, logStart);
+        assertEquals(parsedLogcat.size(), 2);
```
muhomorr: This assert fails on my device sometimes because list size is 1 instead of 2.

Also, expected value should be the first argument, it was a mistake on my part to suggest the current order of arguments in a previous comment.
(failure message says "expected:<1> but was:<2>", should be the other way around)

muhomorr (changes requested):

tests/GrapheneOS/ConnectivityCheckSettingTest/src/grapheneos/test/ConnectivityCheckSettingTest.java
```diff
@@ -0,0 +1,159 @@
...
+        assertTrue(getDevice().setProperty(SYS_PROP, Integer.toString(setting)));
+        assertTrue(toggleWifi("enabled"));
+        boolean isEnabled = false;
+        for (int i = 0; i < 100; ++i) {
+            RunUtil.getDefault().sleep(100);
+            isEnabled = getDevice().executeShellV2Command("cmd wifi status").toString().contains("Wifi is connected to");
+            if (isEnabled) {
+                break;
+            }
+        }
```
muhomorr, line 120: Missing assertTrue for isEnabled

tests/GrapheneOS/ConnectivityCheckSettingTest/src/grapheneos/test/ConnectivityCheckSettingTest.java
```diff
@@ -0,0 +1,162 @@
...
+        assertTrue(isEnabled);
+        return getDevice().getDeviceDate();
```
muhomorr, line 122: Need to capture date before calling `toggleWifi("enabled")` to avoid a race condition.

tests/GrapheneOS/ConnectivityCheckSettingTest/src/grapheneos/test/ConnectivityCheckSettingTest.java
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    private final boolean checkLogSize(String regexp, int setting, long logStart, int logSize) throws Exception {
+        boolean isCorrectSize = false;
+        for (int i = 0; i < 20; ++i) {
+            RunUtil.getDefault().sleep(500);
+            isCorrectSize = (parseLogcat(regexp, setting, logStart).size() == logSize);
+            if (isCorrectSize) {
+                break;
```
muhomorr, line 131: Better to return parseLogcat() result from this function directly (when its size is correct) I think.

PR #447, opened 2023-11-24
Allow backup and restore all Apps

Svirusx:
This changes allow to backup all apps with Android adb backup.

#Backup:
adb backup -apk com.app.name -f app.ab
#Restore:
adb restore app.ab


thestinger:
This violates the security model of production builds. The proper approach is to make a new backup app to replace Seedvault which uses the device-to-device backup mode.

PR #463, opened 2023-12-17
disable screencapture detection api

nycex:
see https://github.com/GrapheneOS/os-issue-tracker/issues/2590


quh4gko8:
See https://github.com/GrapheneOS/os-issue-tracker/issues/2590#issuecomment-1868214967 for plans to handle screenshot privacy concerns. It is preferrable to disable the screenshot detection API instead

PR #498, opened 2024-03-13
bugfix: properly acquire package state to avoid crash from update in original package

quh4gko8:


services/core/java/com/android/server/pm/InstallPackageHelper.java
```diff
@@ -1263,7 +1263,11 @@ private void preparePackageLI(InstallRequest request) throws PrepareFailure {
         boolean systemApp = false;
         boolean replace = false;
         synchronized (mPm.mLock) {
-            final PackageSetting ps = mPm.mSettings.getPackageLPr(pkgName);
+            // See Computer#getPackageStateInternal, which resolves interal package name
+            // before getting PackageSetting, implementation of PackageStateInternal.
+            final String internalPkgName = mPm.snapshotComputer().resolveInternalPackageName(
```
muhomorr: `resolveInternalPackageName` is probably not the right method to use here, it's not used by any other method in this class

services/core/java/com/android/server/pm/InstallPackageHelper.java
```diff
@@ -1263,7 +1263,11 @@ private void preparePackageLI(InstallRequest request) throws PrepareFailure {
         boolean systemApp = false;
         boolean replace = false;
         synchronized (mPm.mLock) {
-            final PackageSetting ps = mPm.mSettings.getPackageLPr(pkgName);
+            // See Computer#getPackageStateInternal, which resolves interal package name
```
muhomorr: Typo

PR #503, opened 2024-03-23
Feat clipboard access

octocorvus:
This is an initial implementation for denying clipboard read access on a per-app basis. As of now, there isn't a way to explicitly paste clipboard content, so this feature is disabled by default, and can be turned on using the clipboard setting in the privacy section.

Depends On:
https://github.com/GrapheneOS/platform_packages_apps_Settings/pull/236


services/core/java/com/android/server/clipboard/ClipboardService.java
```diff
@@ -1472,6 +1492,18 @@ private void showAccessNotificationLocked(String callingPackage, int uid, @UserI
                 clipboard.deviceId) == uid) {
             return;
         }
+
+        // Shows a system notification instead, when clipboard access is blocked.
+        if (isReadBlocked) {
+            String message =
+                    getContext().getString(R.string.notif_clipboard_read_deny_title, callingPackage);
```
muhomorr: Resource strings are usually not used for logging due to their complexity: they usually vary with device locale, retrieving them is slow, etc

services/core/java/com/android/server/clipboard/ClipboardService.java
```diff
@@ -1615,4 +1647,9 @@ private TextClassificationManager createTextClassificationManagerAsUser(@UserIdI
         Context context = getContext().createContextAsUser(UserHandle.of(userId), /* flags= */ 0);
         return context.getSystemService(TextClassificationManager.class);
     }
+
+    private boolean isReadBlockedForPkg(int uid, String pkg, int userId, Clipboard clipboard) {
+        return !UserHandle.isSameApp(uid, clipboard.primaryClipUid)
```
muhomorr: isSameApp check implies that app should have access to its own clipboard contents across user profiles, which is blocked elsewhere

PR #521, opened 2024-05-22
increase default max password length to 128

### LINKED ISSUE #3540: Raise max password length from 64 characters to 128 characters
GrapheneOS currently raises the max length for passwords to 64 characters. While that is ample for random passwords, people are more likely going to want to use diceware passphrases, as those are easier to remember. More specifically, they might want to use the EFF large wordlist which specifically chooses to use longer words on average to improve memorability of the list's words.

The longest word in that wordlist seems to be 9 characters. Assuming someone rolls the max, and assuming there's one space between each words, that's 99 characters for a 10 word passphrase using that list.

Going for 128 characters is clean (double of 64), and gives enough play for people to go even beyond 10 words without issues.

neelchauhan:
Closes https://github.com/GrapheneOS/os-issue-tracker/issues/3540

### DIFF
--- a/core/java/android/app/admin/DevicePolicyManager.java
+++ b/core/java/android/app/admin/DevicePolicyManager.java
@@ -3634,7 +3634,7 @@ public class DevicePolicyManager {
      * Maximum supported password length. Kind-of arbitrary.
      * @hide
      */
-    public static final int MAX_PASSWORD_LENGTH = 64;
+    public static final int MAX_PASSWORD_LENGTH = 128;
 
     /**
      * Service Action: Service implemented by a device owner or profile owner supervision app to


core/java/android/app/admin/DevicePolicyManager.java
```diff
@@ -17168,4 +17168,4 @@ public String getFinancedDeviceKioskRoleHolder() {
     public boolean isOnboardingBugreportV2FlagEnabled() {
         return onboardingBugreportV2Enabled();
     }
-}
```
quh4gko8: Code style nit: this portion of the change needs to be removed and only the diff that should be shown is the change of `MAX_PASSWORD_LENGTH`

quh4gko8: Refer to https://github.com/GrapheneOS/platform_frameworks_base/commit/e9e3a3febbaa7a8ac20bce937f8fdc3d1820a81c on the expected diff

neelchauhan:
Sorry for the delay. I had to upgrade my laptop's RAM in order to build as I cannot run emulator on my GPU-less server and my desktop is packed.

I have some good news:

 * Yes, this patch does work
 * I also removed the newline from the PR

Demo: https://youtu.be/UugvPpYNwIM (The password being entered is on the right of the video in the terminal).

core/java/android/app/admin/DevicePolicyManager.java
neelchauhan: Done!

matchboxbananasynergy:
Thank you. We need to make sure that using the max (128 characters) works fine for both the owner and secondary profiles, that the phone can be unlocked just fine etc.

PR #522, opened 2024-05-24
Use a general message when a PIN is required after booting or after some time

### LINKED ISSUE #2607: [Quickwin] Use general message for a required PIN
Currently, there are three cases where you need to enter your PIN:

* After booting into the system
* After some time (for extra security)
* When in lockdown

The message that a PIN is required is different in each of these scenarios. We shall not distinguish between them because this leaks information. This leaks the information if a user has voluntarily enabled lockdown mode. This can be used against a user.

I propose a message like "PIN required" or "PIN required for security" or "PIN required after boot / after lockdown or for extra security"

Alternatively add a option to make the message generic.

neelchauhan:
This prevents certain information from being leaked if an attacker gets their hands on a phone.

Closes https://github.com/GrapheneOS/os-issue-tracker/issues/2607


matchboxbananasynergy:
Comments say that this also applies to pattern (doesn't matter since we disable it) and passwords. Is that true? If so, I don't think it makes sense for the message to just say "PIN"

neelchauhan:
I noticed that too. Updated the PR for this reason.

matchboxbananasynergy:
Thanks. Unsure if we want to change the pattern-specific strings as we don't use pattern unlock on GrapheneOS. Will defer to the development team on that.

Have you tested and confirmed the correct strings appear in all situations?

neelchauhan:
I'm building GrapheneOS right now. I will upload a demo video later today with the new strings

matchboxbananasynergy:
What do you think about adjusting https://github.com/GrapheneOS/platform_frameworks_base/pull/522/files#diff-bf74eec0499c25852b2af702a7988a992fbf66ed4522037b57086df78159ee89R298-L305?

neelchauhan:
Done!

Also, demo video: https://www.youtube.com/shorts/cGeHMyRIvaw

quh4gko8:
This current patchset doesn't work when other language is set and doesn't resolve the linked issue.

quh4gko8:
An automated way to replace the string name referencing the original "Enter your pin", such as `keyguard_enter_your_pin` would be preferable to resolve the issue across all translations. See Vanadium string rebranding patch as reference for this

PR #523, opened 2024-05-27
Disallow the camera shortcut toggle from the lockscreen

### LINKED ISSUE #2508: Camera shortcut ignores toggle to not allow camera access when the device is locked.
Settings -> Security -> Allow camera access when the device is locked

neelchauhan:
Closes https://github.com/GrapheneOS/os-issue-tracker/issues/2508

Demo: https://youtube.com/shorts/pi8nH4sXaRA?feature=share

### DIFF
--- a/packages/SystemUI/src/com/android/systemui/camera/CameraGestureHelper.kt
+++ b/packages/SystemUI/src/com/android/systemui/camera/CameraGestureHelper.kt
@@ -24,6 +24,7 @@ import android.content.Context
 import android.content.Intent
 import android.content.pm.PackageManager
 import android.content.pm.ResolveInfo
+import android.ext.settings.ExtSettings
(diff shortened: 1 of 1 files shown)


neelchauhan:
As a note, it seems we can't just use `return` because in Kotlin `return` just exits the loop as opposed to exiting the whole function.

**EDIT:** Nevermind, I was wrong. I got it to work without having to enclose a whole bunch of code. Sorry for so many force pushes.

thestinger:
@neelchauhan It may be better to disable and hide the camera shortcut completely instead of having it still appear but do nothing. Need to think about it.

thestinger:
Merging this since it's a clear improvement and we can consider changing it, but I'm not sure if it makes sense to do that yet.

PR #559, opened 2024-07-31
add optional installation of play apps into work profile

mkg20001:
~~Just a rudimentary PR for now (code works already). Waiting for feedback if change is wanted.~~ ~~with https://github.com/GrapheneOS/os-issue-tracker/issues/1821 arround I'm just assuming yes~~

testing instructions: https://github.com/GrapheneOS/platform_frameworks_base/pull/559#issuecomment-2267344940

Problem:

Currently when using any DPC app (like Google Device Policy or others) that uses play services in any way it expects it to be present in the newly created work profile, as this is the case on phones with un-sandboxed-gapps.

Solution:

This can be solved with installing the play services into the work profile.
Because this needs to happen at the earliest possible moments in order for the DPC app to do it's job properly (mainly Google Device Policy, prob others aswell), this needs to be part of the profile creation itself.

---

~~It is planned to make this optional and disabled by default, as not everyone needs that behaviour.~~ Automatically detects if DPC app uses play services.

~~The UI toggle would be added in GmsCompat.~~

~~Not sure whether to put it's value into GmsCompat, since I'm not sure how easy it is to access that value from DevicePolicyManagerService.~~


mkg20001:
So I tried setting up a regular google work account

There are some other issues specifically with that:
- [x] when play store is installed, it needs to be given the permission to install unknown apps at install time. as otherwise the DPC might disallow installing third-party apps and therefore block the play store from functioning
  - ~~This should be doable as part of the app installation in device policy~~
  - added permission grant via app ops service
- [ ] when adding the account first as regular user, play uses some special intent to start deployment and/or copy the authorization into the work profile. that intent breaks as it requires a special permission.
  - Maybe fixable via GmsCompat magic
  - This isn't really an issue as setup can still be completed, it just requires to login in twice

My profile is also stuck in a weird state as I'm using an emulator (as I don't have a test device) and the work profile needs to be custom-rom-certified seperatly from the main profile which breaks the setup (google.com/android/uncertified)

mkg20001:
I got myself a dev phone, it should arrive next week. Then I can test this properly.

services/devicepolicy/java/com/android/server/devicepolicy/DevicePolicyManagerService.java
```diff
@@ -21241,6 +21244,53 @@ private void maybeInstallDevicePolicyManagementRoleHolderInUser(int targetUserId
...
+        // TODO: Add option to toggle whether or not play should be installed
+        // TODO: possibly copy permissions from existing install in managing user?
+        // TODO: handle play services not being installed correctly (shouldn't do anything in that case)
+        Slogf.d(LOG_TAG, "Installing play for user " + targetUserId);
+
+        List<String> playPkgList = Arrays.asList(PackageId.GSF_NAME, PackageId.GMS_CORE_NAME, PackageId.PLAY_STORE_NAME);
+
+        try {
+            for (final String playPkg : playPkgList) {
+                if (mIPackageManager.isPackageAvailable(playPkg, targetUserId)) {
```
mkg20001, line 21262: not sure if this check is superfluos, copied it from maybeInstallDevicePolicyManagementRoleHolderInUser

services/devicepolicy/java/com/android/server/devicepolicy/DevicePolicyManagerService.java
```diff
@@ -21241,6 +21244,53 @@ private void maybeInstallDevicePolicyManagementRoleHolderInUser(int targetUserId
...
+                            + playPkg + " is already installed in "
+                            + "user " + targetUserId);
+                    return;
+                }
+                Slogf.d(LOG_TAG, "Installing play package "
+                      + playPkg + " in user " + targetUserId);
+                mIPackageManager.installExistingPackageAsUser(
+                    playPkg,
+                    targetUserId,
+                    PackageManager.INSTALL_ALL_WHITELIST_RESTRICTED_PERMISSIONS,
```
mkg20001, line 21273: I don't think I need those, will remove later

services/devicepolicy/java/com/android/server/devicepolicy/DevicePolicyManagerService.java
```diff
@@ -581,7 +581,7 @@ public class DevicePolicyManagerService extends IDevicePolicyManager.Stub {
 
     private static final String ATTRIBUTION_TAG = "DevicePolicyManagerService";
 
-    static final boolean VERBOSE_LOG = false; // DO NOT SUBMIT WITH TRUE
+    static final boolean VERBOSE_LOG = true; // DO NOT SUBMIT WITH TRUE
```
quh4gko8, line 584: Need to remove this change and squash the newer changes once cleaned up.

services/devicepolicy/java/com/android/server/devicepolicy/DevicePolicyManagerService.java
```diff
@@ -21241,6 +21244,53 @@ private void maybeInstallDevicePolicyManagementRoleHolderInUser(int targetUserId
...
+
+            Slogf.d(LOG_TAG, "Granting REQUEST_INSTALL_PACKAGES to Play Store");
+
+            // We need to grant Play Store "Allow from source" as this is not possible
+            // later if DPC policy blocks user installing apps
+            // If DPC policy blocks this setting later the value in UI will be "Enabled by admin"
+
+            final int vendingUID = mInjector.getPackageManagerInternal().getPackageUid(
+                    PackageId.PLAY_STORE_NAME, 0, targetUserId);
+            mInjector.getAppOpsManager().setMode(AppOpsManager.OP_REQUEST_INSTALL_PACKAGES, vendingUID,
```
mkg20001, line 21286: this can be improved like in `        for (AppOpsManager.PackageOps pkgOps : allVpnOps) {` using packageops helper class, to not directly call getPackageManagerInternal

mkg20001:
Ok so if DPC blocks "allow from source" it also blocks all app installs, even from play store as that isn't a system app. There would need to be a patch for that.

mkg20001:
That is now fixed aswell

mkg20001:
What should be done when play in work profile tries to downgrade app? Currenty it fails but it could also be made to just clone the existing version of the app into work profile silently

mkg20001:
Testing with Google managed work profile:

_I made a test google workspace for testing this, if you need an account for testing I can give you access_

- open app with google account support (drive, gmail, etc) that is NOT play store

- add work account: click profile (top right) -> "add another account"

- will crash after sign-in, then return to app

- wait. if nothing happens check if the work account is selected in account chooser (top right) and select

- asks for android device policy to be installed, install

- will prompt to create work profile, click accept & continue

- for emulator:
  - run: `adb shell 'sqlite3 /data/user/$(pm list users | grep Work | grep -o "[0-9]*" | head -n 1)/*/*/gservices.db "select * from main where name = \"android_id\";"'`
  - if command doesn't work wait
  - go to google.com/android/uncertified, add id
  - click on work profile notification to continue setup if you get a "device not certified" error

- click next in work profile setup

- will now ask to login again, do it

- will now say to allow device, if necesarry, for the google workspace (ask your company helpdesk)
  - DO NOT REBOOT DURING THIS IT BREAKS SETUP
  - use ":" menu -> sync policies to check if authorized
  - use notification to go back to work profile setup

- after everything is done prompt should auto-close and work account should be gone from user profile

last steps:
- enable notifications for play in work profile (see notifications from GmsCompat with work icon)
- open play and manually confirm 1) play store self-update 2) all the installs for work apps
- if provisioning apps (first install of everything) doesnt start grant contacts permission

packages/PackageInstaller/src/com/android/packageinstaller/InstallStart.java
```diff
@@ -152,7 +153,15 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
             mAbortInstall = true;
         }
 
-        checkDevicePolicyRestrictions(isTrustedSource);
+        // When DPC disabled user app installation, it restricts the play store aswell
+        // We work arround this by skipping devicePolicyChecks for play store
+        // This needs to be done most likely globally, as main profile may also be restricted
+
+        if (callingPackage.equals(PackageId.PLAY_STORE_NAME)) {
```
mkg20001: alternative is to make play store a trusted source but I think that's not a good idea

quh4gko8: This is incorrect, it's not supposed to unconditionally bypass device policy restrictions on Play store. Having this necessary in the first place puts a question on the current proposed solution, this breaks the current expectations of sandboxed Play apps.

mkg20001: The other case where this might apply is a org owned device, it may also have device policy restrictions on unknown sources.

We can also test for this case properly.

Do you have any suggestions on how to improve this?

REQUEST_INSTALL_PACKAGES is already checked before I think, so this is not bypassed.

mkg20001: One idea is `checkDevicePolicyRestrictions(isTrustedSource || callingPackage.equals(PackageId.PLAY_STORE_NAME));`

That way we skip DISALLOW_INSTALL_UNKNOWN_SOURCES check for play, but not sure if that's enough

mkg20001: Other idea would to add toogle to GmsCompat for this behaviour. Would be more complicated though. (Installing package for all users would need to check toggle value on all users)

quh4gko8: Showing the prompt via Apps to install Play apps in managed profiles when it's being created is a better approach, by calling it via ContentProvider, for example, l the device management app. More detailed discussion for the desired approach for this one is better discussed with @muhomorr , the primary maintainer of sandboxed Play apps compatibility layer

mkg20001:
@flawedworld As requested the crash from setup: 
[Error in Google Play services 370b41359e04.txt](https://github.com/user-attachments/files/16488510/Error.in.Google.Play.services.370b41359e04.txt)

flawedworld:
This can maybe be shimmed in GmsCompat.

mkg20001:
There is also the non-privileged version of this intent, which is what shelter app uses. I think it's mostly the same except the user needs to confirm it. It could be made to use that instead.

mkg20001:
Setup video

https://github.com/user-attachments/assets/a87d24f1-3708-4c5d-81b4-96bc6aec3f13

packages/PackageInstaller/src/com/android/packageinstaller/InstallStart.java
mkg20001: Ok. I pm'd him.

For device policy checks: This also works

```java
 checkDevicePolicyRestrictions(isTrustedSource || callingPackage.equals(PackageId.PLAY_STORE_NAME));
````
This would be less invasive then currently skipping everything related to device policy and doesn't require making play a trusted source.
Unless you have something better.

Comment (quh4gko8) [reply packages/PackageInstaller/src/com/android/packageinstaller/InstallStart.java]: Installing via Apps/App Store would be better, it actually handles the INSTALL_PACKAGES permission, and can be prompted upon profile creation. Furthermore, the current proposed change also breaks the expectations of sandboxed Play apps of being unable to do what unprivileged user apps can't. This also serves as a recurring bypass to checking REQUEST_INSTALL_PACKAGES instead of only having the device policy restriction bypassed on creation only, if necessary.

Comment (mkg20001) [reply packages/PackageInstaller/src/com/android/packageinstaller/InstallStart.java]: > Installing via Apps/App Store would be better, it actually handles the INSTALL_PACKAGES permission, and can be prompted upon profile creation. 

So you mean instead of installExistingPackagesAsUser I would go send intent to app store and then it would install play services during profile creation?

Comment (quh4gko8) [reply packages/PackageInstaller/src/com/android/packageinstaller/InstallStart.java]: It's one of the possible approach to be taken rather than putting bypasses for sandboxed Play apps just for it to be installed during the profile creation.

Comment (mkg20001) [reply packages/PackageInstaller/src/com/android/packageinstaller/InstallStart.java]: > It's one of the possible approach to be taken rather than putting bypasses for sandboxed Play apps just for it to be installed during the profile creation.

This is not during profile creation, that's already handled by maybeInstallPlay in devicepolicymanagerservice.

The changes to installstart are because of Device policy blocking installs from play otherwise. Because REQUEST_INSTALL_PACKAGES permission is not enough when DISALLOW_INSTALL_UNKNOWN_SOURCES is set (as is common nowadays with work profiles). Because play is not a system app this restriction would otherwise cause "Blocked by device admin" when installing apps from play in work profile.

**Comment by muhomorr** [comment]
I think this has to be addressed in a different way.

Work profile manager has to install the Google Play trio (GSF, GmsCore, Play Store) itself, it can be made to do so as part of work profile creation. DISALLOW_INSTALL_UNKNOWN_SOURCES issue can be fixed by adding a device policy API for allowlisting specific third-party package installers and making the work profile manager use this API to allowlist the Play Store.

**Comment by mkg20001** [comment]
> DISALLOW_INSTALL_UNKNOWN_SOURCES issue can be fixed by adding a device policy API for allowlisting specific third-party package installers and making the work profile manager use this API to allowlist the Play Store.

Makes sense, will do

> Work profile manager has to install the Google Play trio (GSF, GmsCore, Play Store) itself, it can be made to do so as part of work profile creation.

What do you mean by "Work profile manager"? DevicePolicyManagerService or some other part? DevicePolicyManagerService currently just copies the package over with installPackageAsExistingUser, so you want it to instead also install play services if they are not already installed?

**Comment by muhomorr** [comment]
By "work profile manager" I mean the app that creates and manages the work profile. It has permission to install apps from other users into work profile without confirmation.

**Comment by mkg20001** [comment]
>Work profile manager has to install the Google Play trio (GSF, GmsCore, Play Store) itself, it can be made to do so as part of work profile creation

That is you are saying that all the apps that create work profiles will now need to acount for GOS not delivering play store as system app and having to install it afterwards? They won't. Even if we patch them ourselves similarly to play store, there are countless apps (not just the one from google), it will be easier to let the user decide to install play store during profile creation. Patching all of them is just not realistic.

I already have working code for installing play store into work profile, I will make it a prompt for installing it during work profile creation and add a permission that will be granted by that code to play store (so I don't need to skip checks for allow_unknown_sources). Edit: I can make the permission available to any app in device profile aswell.

Edit: And I understand your concern as to not provide anything special for play store. But unless play store as non-system-app is officially a thing, I don't think work profile manager apps can be required to account for it not being a system app.  We're not giving google any special permissions either way.

**Comment by mkg20001** [comment]
Just so I get it: INSTALL_PACKAGES is not something we should grant to play store, even in work profile and therefore can only use REQUEST_INSTALL_PACKAGES? Correct? Or not?

**Comment by mkg20001** [comment]
I've rebased the changes against latest gos. Also please don't just force push this branch while I'm working on it.

Currently build may be broken as I'm just tossing commits in here for rebasing, etc. I will make a clean history once the major changes are in place.

**Comment by thestinger** [comment]
GrapheneOS is now based on the August release of the Android Open Source Project and was rebased onto it. It's how this always works to keep the project at all maintainable and sustainable. Merging in all the releases would be horrific particularly for quarterly and yearly releases. Keeping it clean leading up to those is incredibly important for moving quickly. We can't stop porting to new releases and preparing for upcoming ones because someone has a PR open. Android 15 is coming soon and there will be major rebases to prepare for it soon.

**Comment by muhomorr** [comment]
> That is you are saying that all the apps that create work profiles will now need to acount for GOS not delivering play store as system app and having to install it afterwards?

This is not what I meant. By "work profile manager can be made to do so as part of work profile creation"
I meant that it can be forced to do so by the OS, without any cooperation from the work profile manager app.

> Even if we patch them ourselves similarly to play store

GSF, Play services, Play Store APKs are not modified in any way on GrapheneOS.

> Patching all of them is just not realistic.

There are reliable ways to detect work profile manager apps attempting to create a work profile,
and a reliable way to detect whether an app depends on Play services, there's a marker in AndroidManifest 
that is added by Play services client library.

**Comment by mkg20001** [comment]
Ok, point taken. That seems reasonable. Let's do that.
Note that I'm not that great with android development yet so I don't think I can get this implemented all by myself.

**Comment by mkg20001** [comment]
I've added a custom permission USER_TRUSTED_SOURCE to allow skipping unknown apps check in work profile (unknown sources restriction) that can be granted to any app so it's not play store specific (need to figure out how to best make granting this permission work, so it doesn't become a security issue)

~~Android seems to not be working with the new permission~~

~~I get errors like these. The common thing is   	at android.app.AppOpsManager.strOpToOp(AppOpsManager.java:8711) being called with null~~

~~It's an appop permission, similar to REQUEST_INSTALL_PACKAGES. Something seems to be missing but I can't figure out what.~~

fixed it

```
 Process: com.android.providers.media.module, PID: 6164
  java.lang.RuntimeException: Unable to get provider com.android.providers.media.MediaProvider: java.lang.IllegalArgumentException: Unknown operation string: null
  	at android.app.ActivityThread.installProvider(ActivityThread.java:8215)
  	at android.app.ActivityThread.installContentProviders(ActivityThread.java:7725)
  	at android.app.ActivityThread.handleBindApplication(ActivityThread.java:7414)
  	at android.app.ActivityThread.-$$Nest$mhandleBindApplication(Unknown Source:0)
  	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2386)
  	at android.os.Handler.dispatchMessage(Handler.java:107)
  	at android.os.Looper.loopOnce(Looper.java:232)
  	at android.os.Looper.loop(Looper.java:317)
  	at android.app.ActivityThread.main(ActivityThread.java:8623)
  	at java.lang.reflect.Method.invoke(Native Method)
  	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:580)
  	at com.android.internal.os.ExecInit.main(ExecInit.java:50)
  	at com.android.internal.os.RuntimeInit.nativeFinishInit(Native Method)
  	at com.android.internal.os.RuntimeInit.main(RuntimeInit.java:369)
  Caused by: java.lang.IllegalArgumentException: Unknown operation string: null
  	at android.app.AppOpsManager.strOpToOp(AppOpsManager.java:8711)
  	at android.app.AppOpsManager.startWatchingMode(AppOpsManager.java:8233)
  	at com.android.providers.media.MediaProvider.onCreate(MediaProvider.java:1447)
  	at android.content.ContentProvider.attachInfo(ContentProvider.java:2644)
  	at android.content.ContentProvider.attachInfo(ContentProvider.java:2613)
  	at com.android.providers.media.MediaProvider.attachInfo(MediaProvider.java:1326)
  	at android.app.ActivityThread.installProvider(ActivityThread.java:8210)
  	at android.app.ActivityThread.installContentProviders(ActivityThread.java:7725) 
  	at android.app.ActivityThread.handleBindApplication(ActivityThread.java:7414) 
  	at android.app.ActivityThread.-$$Nest$mhandleBindApplication(Unknown Source:0) 
  	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2386) 
  	at android.os.Handler.dispatchMessage(Handler.java:107) 
  	at android.os.Looper.loopOnce(Looper.java:232) 
  	at android.os.Looper.loop(Looper.java:317) 
  	at android.app.ActivityThread.main(ActivityThread.java:8623) 
  	at java.lang.reflect.Method.invoke(Native Method) 
  	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:580) 
  	at com.android.internal.os.ExecInit.main(ExecInit.java:50) 
  	at com.android.internal.os.RuntimeInit.nativeFinishInit(Native Method) 
  	at com.android.internal.os.RuntimeInit.main(RuntimeInit.java:369) 
```

mkg20001:
This is now working so far, automatically triggering install of play services on a work profile created by google's android device policy.

What's missing is the logic to detect if the app needs play services for the work profile.

How could this be checked the most easy way?

mkg20001:
Note that for this to build you need framework/proto_logging checked out at github.com/mkg20001/frameworks_proto_logging branch gapps-work-profile

mkg20001:
@thestinger I didn't get what the force push message said at first, I now realize that it's just the 14 branch in main repo, so it's fine. Lol.

packages/PackageInstaller/src/com/android/packageinstaller/InstallStart.java
mkg20001: Replaced by permission USER_TRUSTED_SOURCE

mkg20001:
This is ready for another round of review.

muhomorr:
This PR is still far from using the approach that I've asked for above.

mkg20001:
> This PR is still far from using the approach that I've asked for above.

What's missing? What woud you like to have changed?

muhomorr:
I've wrote about that already:
https://github.com/GrapheneOS/platform_frameworks_base/pull/559#issuecomment-2270778734

mkg20001:
> Work profile manager has to install the Google Play trio (GSF, GmsCore, Play Store) itself, it can be made to do so as part of work profile creation

I do not understand what do you mean by "work profile manager has to do it"

How do I make work profile manager _do that_? Patch? Add hook somewhere? How/Where?

And how is it better to the current method of just doing that as part of DevicePolicyManagerService itself?

>  DISALLOW_INSTALL_UNKNOWN_SOURCES issue can be fixed by adding a device policy API for allowlisting specific third-party package installers and making the work profile manager use this API to allowlist the Play Store.

So you're looking for something like, say, those functions in DevicePolicyManager? Or is there a nicer way without making 3 new functions?

````
@requirepermission(MANAGE_DEVICE_POLICY_RUNTIME_PERMISSIONS)
setTrustedSource(packageName, userId) -> void

@requirepermission(MANAGE_DEVICE_POLICY_RUNTIME_PERMISSIONS)
unsetTrustedSource(packageName, userId) -> void

isTrustedSource(packageName, userId) -> bool
````

Also where would I store the value? Set it per app via appops + USER_TRUSTED_SOURCE permission or is there a better way?

muhomorr:
> How do I make work profile manager do that? Patch? Add hook somewhere? How/Where?

There are several approaches used by the GmsCompat layer to achieve this. It's hard to tell which one would fit best without looking into this in depth.

> And how is it better to the current method of just doing that as part of DevicePolicyManagerService itself?

DevicePolicyManagerService runs inside system_server, which is a highly privileged system component. Introducing a bug there would be a lot more serious than introducing a bug in an unprivileged work profile manager app process. Moreover, system_server internal APIs change often, while the public work profile manager app APIs are very stable, which helps a lot in maintaining such changes.

> So you're looking for something like, say, those functions in DevicePolicyManager? Or is there a nicer way without making 3 new functions?

At a minimum, APK certificate digests should be specified and checked.

> Also where would I store the value? Set it per app via appops + USER_TRUSTED_SOURCE permission or is there a better way?

It's hard to say which approach to use without researching this in-depth. I currently don't have enough time for that.

mkg20001:
Thanks for the response! Yes, makes sense to put it outside of DevicePolicyManagerService.

Do you happen to know somebody in or outside the team that I can ask more in-depth questions that has time for that? Or do you have any guides about android / graphene platform development you could recommend? Or should I just stick to asking in the Development channel?

muhomorr:
> Do you happen to know somebody in or outside the team that I can ask more in-depth questions that has time for that?

No.

> Or do you have any guides about android / graphene platform development you could recommend?

I can recommend https://source.android.com/ and reading the code itself.

turbamulta:
Thank you for the effort trying to get managed work profiles to work. Following.

mkg20001:
Fyi I'm currently busy and will continue this work after android 15 release, as I suspect there might be major changes. If anyone has any hints for me on how to best implement this, just comment.

mkg20001:
new pr in new repo: https://github.com/GrapheneOS/platform_frameworks_base/pull/4

